PR #1, opened 2020-11-16
while-loops: Content

tildeee:
Priority is to review the content (words!)

Examples and questions still need to be filled out.

- Open to suggestions on examples and questions

### DIFF
--- a/while-loops/while-loops.md
+++ b/while-loops/while-loops.md
@@ -2,12 +2,12 @@
 
 ## Learning Goals
 
-- Practice using while loops to loop while a conditional is true 
-- Apply a counter variable while using while loops
+* Practice using while loops to loop while a conditional is true 
+* Apply a counter variable while using while loops
 
 ## Introduction
 
-Imagine writing software that creates digital group greeting cards. Before the user can sign a digital group greeting card, they need to type in the specific password `AdaLovelaceCodesIt`. If they don't type in the correct password, then the app should ask them to type in the password again, and again, and again, until the user leaves or types in `AdaLovelaceCodesIt`.
+Imagine writing software that creates digital group greeting cards. Before the user can sign a digital group greeting card, they need to type in the specific password `AdaLovelaceCodesIt` . If they don't type in the correct password, then the app should ask them to type in the password again, and again, and again, until the user leaves or types in `AdaLovelaceCodesIt` .
 
 How can we solve this problem using Python code?
 
@@ -24,22 +24,22 @@ Counter variable | A variable solely used to representing a number that incremen
 
 While loops are loops that run _while_ some condition is `True` or truthy.
 
-```python
+``` python
 while conditional_evaluates_to_true:
     print("I'm in the loop body")
 ```
 
-| Piece of code | Notes |
---- | ---
-`while` | `while` is a keyword in Python that begins a while loop. It looks immediately to the right for a conditional expression, and a `:` to end the line
-`conditional_evalutes_to_true` | **Replace this** with a conditional expression. When this expression is true, the loop body will run
-Loop body | **Replace this** with the code that should run in this while loop. This should be indented once inside the while loop.
+| Piece of code                  | Notes                                                                                                                                              |
+|--------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------|
+| `while` | `while` is a keyword in Python that begins a while loop. It looks immediately to the right for a conditional expression, and a `:` to end the line |
+| `conditional_evalutes_to_true` | **Replace this** with a conditional expression. When this expression is true, the loop body will run                                               |
+| Loop body                      | **Replace this** with the code that should run in this while loop. This should be indented once inside the while loop.                             |
 
 ### Practicing Reading While Loops
 
 Here, we use a while loop to print the numbers 1 through 5.
 
-```python
+``` python
 i = 1
 while i <= 5:
     print(f"The value of i is {i}")
@@ -48,26 +48,26 @@ while i <= 5:
 print("We are outside the while loop! Wooo!")
 ```
 
-Here, we have a variable `i`. While `i` is less than or equal to 5, we will print `f"The value of i is {i}"`. Then, we will increment `i` by one.
+Here, we have a variable `i` . While `i` is less than or equal to 5, we will print `f"The value of i is {i}"` . Then, we will increment `i` by one.
 
-`i` begins with a value of `1`. As the while loop loops, `i` grows bigger and bigger, since we're modifying the same `i` variable.
+`i` begins with a value of `1` . As the while loop loops, `i` grows bigger and bigger, since we're modifying the same `i` variable.
 
-At some point, `i` will have a value of `6`. Then, `i <= 5` is falsey, and the while loop will exit. Our code execution moves on to the final `print` statement.
+At some point, `i` will have a value of `6` . Then, `i <= 5` is falsey, and the while loop will exit. Our code execution moves on to the final `print` statement.
 
 Another way to read this while loop is with a table:
 
-Value of `i` at the beginning of the loop | Is `i <= 10`? | Value of `i` at the end of the loop body
---- | --- | ---
-`1` | `True` | `2`
-`2` | `True` | `3`
-`3` | `True` | `4`
-`4` | `True` | `5`
-`5` | `True` | `6`
-`6` | `False` | Loop body doesn't execute
+| Value of `i` at the beginning of the loop | Is `i <= 10` ? | Value of `i` at the end of the loop body |
+|-------------------------------------------|---------------|------------------------------------------|
+| `1` | `True` | `2` |
+| `2` | `True` | `3` |
+| `3` | `True` | `4` |
+| `4` | `True` | `5` |
+| `5` | `True` | `6` |
+| `6` | `False` | Loop body doesn't execute                |
 
-Here, we should observe that at some point, `i` has a value of `6`, and the code asked "Is 6 <= 5? Is that true?" Because the answer is false, the code skips over the rest of the loop.
+Here, we should observe that at some point, `i` has a value of `6` , and the code asked "Is 6 <= 5? Is that true?" Because the answer is false, the code skips over the rest of the loop.
 
-Callout: While loops can accidentally run infinitely! If the conditional for the while loop never becomes false, then the while loop and the program keep going, and going, and going... until something forces it to stop (such as a human closing down the program), or our machine cannot run it anymore. We call these cases **infinite loops,** and are always a bug.
+Callout: While loops can accidentally run infinitely! If the conditional for the while loop never becomes false, then the while loop and the program keep going, and going, and going... until something forces it to stop (such as a human closing down the program), or our machine cannot run it anymore. We call these cases **infinite loops, ** and are always a bug.
 
 ## Used While Problem-Solving
 
@@ -75,7 +75,7 @@ Let's explore different examples that apply while loops.
 
 We can use a while loop even if we are decrementing a variable...
 
-```python
+``` python
 print("Preparing for blastoff in")
 
 countdown = 10
@@ -89,7 +89,7 @@ print("Launch!")
 
 We can use a while loop that uses a function to calculate something...
 
-```python
+``` python
 flower_basket = []
 
 while len(flower_basket) < 12:
@@ -98,11 +98,11 @@ while len(flower_basket) < 12:
 print("I have a very full flower basket with a dozen flowers:", flower_basket)
 ```
 
-Callout: Note that this conditional uses `<`, _not_ `<=`, because we don't want to add another flower if the length of `flower_basket` is _equal_ to twelve. Watch out for an off-by-one error!
+Callout: Note that this conditional uses `<` , _not_ `<=` , because we don't want to add another flower if the length of `flower_basket` is _equal_ to twelve. Watch out for an off-by-one error!
 
 We can use a while loop that only checks if a variable is truthy!
 
-```python
+``` python
 import random
 coin_is_tails = True
 
@@ -118,7 +118,7 @@ print("We finally got heads.")
 
 Let's return to our original example:
 
-```python
+``` python
 i = 1
 while i <= 5:
     print(f"The value of i is {i}")
@@ -131,9 +131,9 @@ A variable whose job it is to increment or decrement, specifically in a loop, is
 
 This counter variable is:
 
-- used in the conditional expression for the while loop
-- must eventually be modified in the while loop
-- typically initialized before the loop
+* used in the conditional expression for the while loop
+* must eventually be modified in the while loop
+* typically initialized before the loop
 
 In this case, `i` is the counter variable. It gets setup before the while loop begins, and then increments inside the while loop. When we forget to change our counter variable, we often get infinite loops and bugs!
 
@@ -143,7 +143,7 @@ Callout: An alternative syntax is to use for loops with a range. The for loop sy
 
 `break` is a keyword. When our code comes across `break` in a while loop, it will exit the entire while loop. The next line that runs is the line after the while loop.
 
-```python
+``` python
 def demonstrate_break():
     i = 3
 
@@ -158,12 +158,13 @@ demonstrate_break()
 ```
 
 Possible example use cases:
-- needing to find a specific element, then leave the iteration
-- needing to leave the loop in case it comes across some invalid or error state
 
-`continue` is a keyword that will stop executing the current iteration, and resume execution at the next iteration. When our code sees `continue`, it doesn't run anymore code in that iteration, but continues to the next loop.
+* needing to find a specific element, then leave the iteration
+* needing to leave the loop in case it comes across some invalid or error state
 
-```python
+`continue` is a keyword that will stop executing the current iteration, and resume execution at the next iteration. When our code sees `continue` , it doesn't run anymore code in that iteration, but continues to the next loop.
+
+``` python
 def demonstrate_continue():
     i = 3
 
@@ -180,7 +181,8 @@ demonstrate_continue()
 ```
 
 Possible example uses cases:
-- wanting a loop to "skip" in a special case
+
+* wanting a loop to "skip" in a special case
 
 ## Check for Understanding
 
@@ -188,9 +190,9 @@ Possible example uses cases:
 
 <!-- question stubs -->
 
-Fix this loop so it uses the counter variable `i`, and runs while `i` is less than `5`.
+Fix this loop so it uses the counter variable `i` , and runs while `i` is less than `5` .
 
-```python
+``` python
 def loop_until_five():
     i = 0
 
@@ -206,7 +208,7 @@ result = loop_until_five()
 
 Which of these loop table options most accurately describes this loop?
 
-```python
+``` python
 i = 0
 result = None
 
@@ -219,19 +221,19 @@ while i < 5:
 ```
 
 <!-- Answer: This is the correct answer. False answers should toggle True/False on result, or be off-by-one on the i increments -->
-Value of `i` at the beginning of the loop | `i < 5`? | `i` at the end of loop body | `result` at end of loop body
---- | --- | --- | ---
-`0` | `True` | `1` | `True`
-`1` | `True` | `2` | `False`
-`2` | `True` | `3` | `True`
-`3` | `True` | `4` | `False`
-`4` | `True` | `5` | `True`
-`5` | `False` | Loop body doesn't execute | -
 
+| Value of `i` at the beginning of the loop | `i < 5` ? | `i` at the end of loop body | `result` at end of loop body |
+|-------------------------------------------|----------|-----------------------------|------------------------------|
+| `0` | `True` | `1` | `True` |
+| `1` | `True` | `2` | `False` |
+| `2` | `True` | `3` | `True` |
+| `3` | `True` | `4` | `False` |
+| `4` | `True` | `5` | `True` |
+| `5` | `False` | Loop body doesn't execute   | -                            |
 
 Read this code. The function `look_for_treasure` should exit after we find treasure. Add in the `break` keyword in the right line to make this happen and pass the test.
 
-```python
+``` python
 import random
 
 def check_if_treasure(item):
@@ -253,4 +255,3 @@ result = look_for_treasure()
 ```
 
 <!-- Answer: i should be "Treasure" -->
-
--- a/while-loops/while-loops.md
+++ b/while-loops/while-loops.md
@@ -1,0 +1,256 @@
+# While Loops
+
+## Learning Goals
+
+- Practice using while loops to loop while a conditional is true 
+- Apply a counter variable while using while loops
+
+## Introduction
+
+Imagine writing software that creates digital group greeting cards. Before the user can sign a digital group greeting card, they need to type in the specific password `AdaLovelaceCodesIt`. If they don't type in the correct password, then the app should ask them to type in the password again, and again, and again, until the user leaves or types in `AdaLovelaceCodesIt`.
+
+How can we solve this problem using Python code?
+
+for loops allow programmers to iterate over lists. Sometimes, we need to repeat lines of code, but not because we are iterating through a list. We can explore the while loop as an alternative!
+
+## Vocabulary and Synonyms
+
+| Vocab | Definition | Synonyms | How to Use in a Sentence
+| --- | --- | --- | ---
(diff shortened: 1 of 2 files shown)


while-loops/while-loops.md
```diff
@@ -0,0 +1,257 @@
...
+| Value of `i` at the beginning of the loop | Is `i <= 10` ? | Value of `i` at the end of the loop body |
+|-------------------------------------------|---------------|------------------------------------------|
+| `1` | `True` | `2` |
+| `2` | `True` | `3` |
+| `3` | `True` | `4` |
+| `4` | `True` | `5` |
+| `5` | `True` | `6` |
+| `6` | `False` | Loop body doesn't execute                |
+
+Here, we should observe that at some point, `i` has a value of `6` , and the code asked "Is 6 <= 5? Is that true?" Because the answer is false, the code skips over the rest of the loop.
```
bion, line 68: > Because the answer is false, the code skips over the rest of the loop.

Consider rewording to "Because the answer is false, the program does not execute the loop body again and exits the loop"

while-loops/while-loops.md
```diff
@@ -0,0 +1,257 @@
...
+
+Callout: While loops can accidentally run infinitely! If the conditional for the while loop never becomes false, then the while loop and the program keep going, and going, and going... until something forces it to stop (such as a human closing down the program), or our machine cannot run it anymore. We call these cases **infinite loops, ** and are always a bug.
```
bion, line 70: Markdown formatting for bolding isn't working, maybe should be `**infinite loops**,` with the comma outside?

> and are always a bug

Recommend rewording to "and are often a bug".

There are plenty of examples of programs with infinite loops that are not bugs. Event loop based concurrency is often implemented with an explicit `while True:` loop. E.g. https://tutorialedge.net/python/concurrency/asyncio-event-loops-tutorial/

while-loops/while-loops.md
```diff
@@ -0,0 +1,257 @@
...
+
+## Used While Problem-Solving
```
bion, line 72: I'm not sure about this header? Maybe "Problem-Solving using `while`"?

while-loops/while-loops.md
```diff
@@ -0,0 +1,257 @@
...
+``` python
+i = 1
+while i <= 5:
+    print(f"The value of i is {i}")
+    i += 1
+
+print("We are outside the while loop! Wooo!")
+```
+
+A variable whose job it is to increment or decrement, specifically in a loop, is often called a _counter variable._
```
bion, line 130: Nit: consider rewording to "A variable that is only incremented or decremented, ...". Saying that a variable _does_ something sounds a little strange.

while-loops/while-loops.md
```diff
@@ -0,0 +1,257 @@
...
+
+This counter variable is:
```
bion, line 132: s/This/A

while-loops/while-loops.md
```diff
@@ -0,0 +1,257 @@
...
+
+* used in the conditional expression for the while loop
+* must eventually be modified in the while loop
+* typically initialized before the loop
+
+In this case, `i` is the counter variable. It gets setup before the while loop begins, and then increments inside the while loop. When we forget to change our counter variable, we often get infinite loops and bugs!
+
+Callout: An alternative syntax is to use for loops with a range. The for loop syntax sets up and modifies something like a counter variable!
```
bion, line 140: recommend code formatting for "`for` loops" ... "The `for` loop"

while-loops/while-loops.md
```diff
@@ -0,0 +1,257 @@
...
+    i = 3
+
+    while i < 10:
+        if i == 7:
+            print("i is seven! Let's get out of here!")
+            break
+        print(f"Counting: i is {i}")
+        i += 1
+
+demonstrate_break()
```
bion, line 157: I'm not sure the wrapper function helps the immediate learning goal since the student may think the function is somehow relevant to the `break` keyword?

while-loops/while-loops.md
```diff
@@ -0,0 +1,257 @@
...
+```
+
+Possible example use cases:
+
+* needing to find a specific element, then leave the iteration
+* needing to leave the loop in case it comes across some invalid or error state
+
+`continue` is a keyword that will stop executing the current iteration, and resume execution at the next iteration. When our code sees `continue` , it doesn't run anymore code in that iteration, but continues to the next loop.
```
bion, line 165: s/anymore/any more

> but continues to the next loop

Recommend rewording to "but continues to the next iteration of the loop".

Making a clear distinction between "loop" as a reference to the overall control structure and "loop iteration" or "iteration of a loop" is helpful I think.

while-loops/while-loops.md
```diff
@@ -0,0 +1,257 @@
...
+
+``` python
+def demonstrate_continue():
```
bion, line 168: Same comment as with the other wrapper function.

Consider explicitly stating that "Counting: i is 7" is never printed.

bion (approved):

while-loops/while-loops.md
```diff
@@ -0,0 +1,257 @@
...
+
+## Introduction
+
+Imagine writing software that creates digital group greeting cards. Before the user can sign a digital group greeting card, they need to type in the specific password `AdaLovelaceCodesIt` . If they don't type in the correct password, then the app should ask them to type in the password again, and again, and again, until the user leaves or types in `AdaLovelaceCodesIt` .
+
+How can we solve this problem using Python code?
+
+for loops allow programmers to iterate over lists. Sometimes, we need to repeat lines of code, but not because we are iterating through a list. We can explore the while loop as an alternative!
+
+## Vocabulary and Synonyms
```
bion, line 16: "terminology" and "term" tend to be more often used to refer to words and phrases in the context of a particular field e.g. computer science.

while-loops/while-loops.md
```diff
@@ -0,0 +1,257 @@
...
+print("We are outside the while loop! Wooo!")
+```
+
+A variable whose job it is to increment or decrement, specifically in a loop, is often called a _counter variable._
+
+This counter variable is:
+
+* used in the conditional expression for the while loop
+* must eventually be modified in the while loop
+* typically initialized before the loop
```
jkepps, line 136: my python-specific knowledge about variable scope is limited, but shouldn't this say "must be initialized before the loop" if we're specifically talking about counter variables in relation to `while` loops?

while-loops/while-loops.md
jkepps: this should apply to everywhere a keyword is used

while-loops/while-loops.md
jkepps: also, it might be helpful to write out the output of this loop (or maybe save it as an exercise).

while-loops/while-loops.md
jkepps: another opportunity to use this as exercise: "What is the output of this `while` loop?"

while-loops/while-loops.md
```diff
@@ -0,0 +1,257 @@
...
+            print("*~*~*~* \o/ *~*~*~*")
+            continue
+        print(f"Counting: i is {i}")
+
+demonstrate_continue()
+```
+
+Possible example uses cases:
+
+* wanting a loop to "skip" in a special case
```
jkepps, line 185: maybe be more explicit about what's being "skipped". Not necessarily skipping an entire iteration, but just certain lines of code in that iteration.

jkepps (approved):

while-loops/while-loops.md
```diff
@@ -0,0 +1,257 @@
...
+
+    while False:
+        i += 1
+
+    return i
+
+result = loop_until_five()
+```
+
+<!-- Answer: i should be 5 -->
```
audreyandoy, line 207:
```suggestion
### !challenge

* type: multiple-choice
* id: 378b2b65-9fe1-4d47-b3af-6776bb1db85c
* title: Using Counter Variables


##### !question

Fix this while loop so it uses the counter variable `i` and loops 5 times.

``` python
def loop_until_five():
    i = 0

    while False:
        i += 1

    return i

result = loop_until_five()
```

##### !end-question

##### !options

* while i < 5:
* while i <= 5:
* while i == 5:
* Loop condition doesn't need to be fixed

##### !end-options

##### !answer

* while i <= 5:

##### !end-answer
### !end-challenge

```

File: while-loops/while-loops.md [hunk 14]
```diff
@@ -0,0 +1,257 @@
...
+<!-- Answer: This is the correct answer. False answers should toggle True/False on result, or be off-by-one on the i increments -->
+
+| Value of `i` at the beginning of the loop | `i < 5` ? | `i` at the end of loop body | `result` at end of loop body |
+|-------------------------------------------|----------|-----------------------------|------------------------------|
+| `0` | `True` | `1` | `True` |
+| `1` | `True` | `2` | `False` |
+| `2` | `True` | `3` | `True` |
+| `3` | `True` | `4` | `False` |
+| `4` | `True` | `5` | `True` |
+| `5` | `False` | Loop body doesn't execute   | -                            |
```
audreyandoy, line 232:
```suggestion
### !challenge

* type: checkbox
* id: f4376535-5a51-4c84-9749-bfe7b9411728
* title: Loop Table

##### !question

Which of these loop table options most accurately describes this loop?

``` python
i = 0
result = None

while i < 5:
    if (i % 2) == 0:
        result = True
    else:
        result = False
    i += 1
```

##### !end-question

##### !options

* 
| Value of `i` at the beginning of the loop | `i < 5` ? | `i` at the end of loop body | `result` at end of loop body |
|-------------------------------------------|----------|-----------------------------|------------------------------|
| `1` | `True` | `1` | `True` |
| `2` | `True` | `2` | `False` |
| `3` | `True` | `3` | `True` |
| `4` | `True` | `4` | `False` |
| `5` | `False` | Loop body doesn't execute   | -                            |

* 
| Value of `i` at the beginning of the loop | `i < 5` ? | `i` at the end of loop body | `result` at end of loop body |
|-------------------------------------------|----------|-----------------------------|------------------------------|
| `1` | `True` | `1` | `True` |
| `2` | `True` | `2` | `False` |
| `3` | `True` | `3` | `True` |
| `4` | `True` | `4` | `False` |
| `5` | `True` | `5` | `True` |
| `6` | `False` | Loop body doesn't execute   | -                            |
* 
| Value of `i` at the beginning of the loop | `i < 5` ? | `i` at the end of loop body | `result` at end of loop body |
|-------------------------------------------|----------|-----------------------------|------------------------------|
| `0` | `True` | `1` | `True` |
| `1` | `True` | `2` | `False` |
| `2` | `True` | `3` | `True` |
| `3` | `True` | `4` | `False` |
| `4` | `True` | `5` | `True` |
| `5` | `False` | Loop body doesn't execute   | -                            |

##### !end-options

##### !answer

* 

| Value of `i` at the beginning of the loop | `i < 5` ? | `i` at the end of loop body | `result` at end of loop body |
|-------------------------------------------|----------|-----------------------------|------------------------------|
| `0` | `True` | `1` | `True` |
| `1` | `True` | `2` | `False` |
| `2` | `True` | `3` | `True` |
| `3` | `True` | `4` | `False` |
| `4` | `True` | `5` | `True` |
| `5` | `False` | Loop body doesn't execute   | -                            |


##### !end-answer

### !end-challenge
```

Comment (audreyandoy) [reply while-loops/while-loops.md:232]: @tildeee Learn doesn't seem to like putting tables as challenge options for me. Is that true for you too? Maybe I'm doing something wrong here.

File: while-loops/while-loops.md [hunk 15]
```diff
@@ -0,0 +1,257 @@
...
+      if i == "Treasure!":
+        print("We found treasure instead of a number!")
+        print("Let's leave while we can.")
+      print(f"Counting: i is {i}")
+      i = check_if_treasure(i)
+
+result = look_for_treasure()
+```
+
+<!-- Answer: i should be "Treasure" -->
```
audreyandoy, line 257:
```suggestion
### !challenge

* type: checkbox
* id: 891a2b93-62bd-4cd6-9b9f-d1be8d7da746
* title: Break keyword


##### !question

Read this code. The function `look_for_treasure` should exit after we find treasure. Select the option that has the `break` keyword in the right line to make this happen.

``` python
import random

def check_if_treasure(item):
    is_treasure = random.randrange(100) < 25
    # The line below uses the ternary operator.
    return "Treasure!" if is_treasure else item + 1

def look_for_treasure():
    i = 0

    while True:
      if i == "Treasure!":
        print("We found treasure instead of a number!")
        print("Let's leave while we can.")
      print(f"Counting: i is {i}")
      i = check_if_treasure(i)

result = look_for_treasure()
```

##### !end-question

##### !options

* 
```python
def check_if_treasure(item):
    is_treasure = random.randrange(100) < 25
    # The line below uses the ternary operator.
    return "Treasure!" if is_treasure else item + 1

def look_for_treasure():
    i = 0

    while True:
      if i == "Treasure!":
        print("We found treasure instead of a number!")
        print("Let's leave while we can.")
        break
      print(f"Counting: i is {i}")
      i = check_if_treasure(i)

result = look_for_treasure()
```

* 
```python
def check_if_treasure(item):
    is_treasure = random.randrange(100) < 25
    # The line below uses the ternary operator.
    return "Treasure!" if is_treasure else item + 1

def look_for_treasure():
    i = 0

    while True:
      if i == "Treasure!":
        print("We found treasure instead of a number!")
        print("Let's leave while we can.")
      print(f"Counting: i is {i}")
      break
      i = check_if_treasure(i)

result = look_for_treasure()
```

*
```python
def check_if_treasure(item):
    is_treasure = random.randrange(100) < 25
    # The line below uses the ternary operator.
    return "Treasure!" if is_treasure else item + 1

def look_for_treasure():
    i = 0

    while True:
      if i == "Treasure!":
        print("We found treasure instead of a number!")
        print("Let's leave while we can.")
      print(f"Counting: i is {i}")
      
      i = check_if_treasure(i)
      break

result = look_for_treasure()
```

##### !end-options

##### !answer

* 
```python
def check_if_treasure(item):
    is_treasure = random.randrange(100) < 25
    # The line below uses the ternary operator.
    return "Treasure!" if is_treasure else item + 1

def look_for_treasure():
    i = 0

    while True:
      if i == "Treasure!":
        print("We found treasure instead of a number!")
        print("Let's leave while we can.")
        break
      print(f"Counting: i is {i}")
      i = check_if_treasure(i)

result = look_for_treasure()
```


##### !end-answer

### !end-challenge

```

**Review by audreyandoy — CHANGES_REQUESTED** [review]
@tildeee feel free to take a look at these suggestions/changes :)

File: while-loops/while-loops.md (hunk 15 above)
Comment (audreyandoy) [reply while-loops/while-loops.md:257]: I also might think this question would be better as part of an activity?

File: while-loops/while-loops.md [hunk 16]
```diff
@@ -0,0 +1,257 @@
+# While Loops
+
+## Learning Goals
+
+* Practice using while loops to loop while a conditional is true 
+* Apply a counter variable while using while loops
```
tildeee, line 6:
```suggestion
* Practice using while loops
* Practice using a counter variable in a while loop
```

while-loops/while-loops.md
```diff
@@ -0,0 +1,257 @@
...
+
+## Introduction
+
+Imagine writing software that creates digital group greeting cards. Before the user can sign a digital group greeting card, they need to type in the specific password `AdaLovelaceCodesIt` . If they don't type in the correct password, then the app should ask them to type in the password again, and again, and again, until the user leaves or types in `AdaLovelaceCodesIt` .
+
+How can we solve this problem using Python code?
+
+for loops allow programmers to iterate over lists. Sometimes, we need to repeat lines of code, but not because we are iterating through a list. We can explore the while loop as an alternative!
```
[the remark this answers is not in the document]
tildeee:
```suggestion
`for` loops allow programmers to iterate over lists. Sometimes, we need to repeat lines of code, but not because we are traversing (iterating through) a list. We can explore the `while` loop as an alternative!
```

Added `for` and `while` formatting -- I will go through these lessons and apply this formatting lesson-wide afterwards

Went with "traversing" as the main verb, and "iterating through" as a parenthetical!

LMK what you think! Closing this suggestion not to stop conversation, but to help me move forward.

while-loops/while-loops.md
tildeee: Good thought-- This header is "good enough" and it's in enough lessons (😆 ) that an update to the header should happen curriculum-wide

while-loops/while-loops.md
```diff
@@ -0,0 +1,257 @@
...
+
+## Vocabulary and Synonyms
+
+| Vocab | Definition | Synonyms | How to Use in a Sentence
+| --- | --- | --- | ---
+Infinite Loop | A loop that runs and never terminates or ends. Leads to a stack overflow error. | endless loop | "The loop kept going and going forever"
+Counter variable | A variable solely used to representing a number that increments or decrements. Usually, it's used to count loops. It needs to be initialized outside of the loop. This is an informal, casual label! | - | "The counter variable `i` started at `0` and went up with each loop"
+
+## While Loops Syntax
```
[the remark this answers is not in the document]
tildeee:
```suggestion
## `while` Loop Syntax
```

Nah, pretty sure that I typo'd and added that extra "s" there, or at least it reads WAY better not-plural

while-loops/while-loops.md
```diff
@@ -0,0 +1,257 @@
...
+
+While loops are loops that run _while_ some condition is `True` or truthy.
```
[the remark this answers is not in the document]
tildeee: Great point!

Truthiness/falsiness is introduced earlier in the curriculum, but this helps me consider making a callout section/small lesson aside to "recall that truthiness/falsiness is..." and reinforce!

I'll add that content in a separate commit

while-loops/while-loops.md
```diff
@@ -0,0 +1,257 @@
...
+while i <= 5:
+    print(f"The value of i is {i}")
+    i += 1
+
+print("We are outside the while loop! Wooo!")
+```
+
+Here, we have a variable `i` . While `i` is less than or equal to 5, we will print `f"The value of i is {i}"` . Then, we will increment `i` by one.
+
+`i` begins with a value of `1` . As the while loop loops, `i` grows bigger and bigger, since we're modifying the same `i` variable.
```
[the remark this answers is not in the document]
tildeee:
```suggestion
`i` begins with a value of `1`. As the `while` loop iterates, `i` grows bigger and bigger, since we're modifying the same `i` variable.
```

while-loops/while-loops.md
tildeee: Awesome suggestion-- I super appreciate the catch on consistency between false/no

```suggestion
Here, we should observe that at some point, `i` has a value of `6` , and the code asked "Is 6 <= 5? Is that true?" Because the answer is no, the program does not execute the loop body again and exits the loop.
```

@jkepps I super agree that "Is 6 <= 5? Is that true?" is redundant... Hm... I personally have this opinion that redundant language can be helpful in this situation, really because "Is 6 <= 5?" is literally a small question in size... but is the big question to slow down on. The redundancy imo could help slow down a reader.

(Also, I'm a believer that a newbie would see "Is 6 <= 5?" and answer "yes," but not immediately connect that to "yes has a direct relationship with true." For example, I'd want a programmer to answer "Is 6 <= 5?" with "true" before "yes" :P ..... if that makes sense....)

I think right now I'll keep the redundancy, but I'm happy for pushback if it's muddying the readability.

while-loops/while-loops.md
tildeee: Thank you for the accuracy callout! I totally agree.

```suggestion
Callout: While loops can accidentally run infinitely! If the conditional for the while loop never becomes false, then the while loop and the program keep going, and going, and going... until something forces it to stop (such as a human closing down the program), or our machine cannot run it anymore. We call these cases **infinite loops,** and are often a bug.
```

Fairly confident the formatting mistake was because of the leading space

while-loops/while-loops.md
tildeee: Lol i literally don't know what i was intending when i wrote that header

while-loops/while-loops.md
```diff
@@ -0,0 +1,257 @@
...
+
+while len(flower_basket) < 12:
+    flower_basket.append("flower")
+
+print("I have a very full flower basket with a dozen flowers:", flower_basket)
+```
+
+Callout: Note that this conditional uses `<` , _not_ `<=` , because we don't want to add another flower if the length of `flower_basket` is _equal_ to twelve. Watch out for an off-by-one error!
+
+We can use a while loop that only checks if a variable is truthy!
```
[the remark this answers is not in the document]
tildeee: This is awesome!

Truthiness/falsiness will be covered earlier, but it definitely needs reinforcement. Hopefully it feels like a reasonable compromise to include both ideas:

```suggestion
We can use a while loop that only checks if a variable is `True` or truthy!
```

Lmk if this doesn't seem like a great compromise

while-loops/while-loops.md
tildeee:
```suggestion
A counter variable is:
```

while-loops/while-loops.md
tildeee: I'll do a sweep in this lesson to add formatting, outside of this PR

Thank you!! I was on the fence about how to stylize `while` loops (still figuring out if I want a hyphen...), so this is good feedback.

while-loops/while-loops.md
tildeee: Re: the wrapper function:

I'm a little torn on this, since teaching at Ada has given me wisdom that showing example code that lives in functions is often helpful
(going on a weird soapbox about this: not having a wrapper function is totally great, but for some reason, at Ada, I often see that students struggle with the idea that programs are made of functions/methods. By this, I mean that for some of the early projects, many students think that they're code shouldn't live in a method, and should all be in this "script" kind of feeling... which, sure, but ends up being a strange habit to callout around week 5)


I just scrolled up and saw that most of the other examples don't have wrapper functions, so I'll remove it for consistency! But I already typed that lovely paragraph above, so it stays. :P

while-loops/while-loops.md
```diff
@@ -0,0 +1,257 @@
...
+    while i < 10:
+        if i == 7:
+            print("i is seven! Let's get out of here!")
+            break
+        print(f"Counting: i is {i}")
+        i += 1
+
+demonstrate_break()
+```
+
```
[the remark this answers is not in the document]
tildeee: I like the suggestion; I'm going to review this lesson visually before I make a call on adding this. Once I get a good visual on it, I'm going to assess if this output box pushes the lesson over into "this is extremely long and my brain is leaky" territory

One reason why this suggestion is in a medium-pri for me is because intro to `for` loops has them do ~20 loops, often with a loop table, so my hope is that this output is only reinforcement. But I def see the value in this output!

while-loops/while-loops.md
tildeee:
```suggestion
* needing to "skip" the rest of a loop body in a special case
```

tildeee:
@audreyandoy Re-requesting a review. Not adding in the Learn md until it's merged in, but I'll use the Learn md comments as reference.

while-loops/while-loops.md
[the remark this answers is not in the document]
jkepps: now that i know truthy/falsy are covered earlier, this is great.

while-loops/while-loops.md
```diff
@@ -0,0 +1,269 @@
...
+| `1` | `True` | `2` |
+| `2` | `True` | `3` |
+| `3` | `True` | `4` |
+| `4` | `True` | `5` |
+| `5` | `True` | `6` |
+| `6` | `False` | Loop body doesn't execute                |
+
+Here, we should observe that at some point, `i` has a value of `6` , and the code asked "Is 6 <= 5? Is that true?" Because the answer is no, the program does not execute the loop body again and exits the loop.
+
+Callout: While loops can accidentally run infinitely! If the conditional for the while loop never becomes false, then the while loop and the program keep going, and going, and going... until something forces it to stop (such as a human closing down the program), or our machine cannot run it anymore. We call these cases **infinite loops,** and are often a bug.
```
audreyandoy, line 70:
```suggestion
<!-- prettier-ignore-start -->
### !callout-danger

## Infinite Loops
While loops can accidentally run infinitely! If the conditional for the while loop never becomes false, then the while loop and the program keep going, and going, and going... until something forces it to stop (such as a human closing down the program), or our machine cannot run it anymore. We call these cases **infinite loops,** and are often a bug.
### !end-callout
<!-- prettier-ignore-end -->


```

audreyandoy (approved):
Looks great so far!

tildeee:
I will merge in all the Learn stuff that @audreyandoy suggested manually!

PR #4, opened 2020-11-19
intro-to-dev-env: computer hardware and binary WIP

tildeee:


tildeee:
Will probably cut this material.

PR #41, opened 2020-12-17
intro-to-tests: code challenges should use unittest, and import *

tildeee:
@audreyandoy the Python code challenges will use unittest, even though we're teaching pytest. Assigning you to this so you have visibility into this kind of big curriculum decision!

I thought through this and decided:
- unittest tests in Learn are more maintainable vs. custom snippets. They are inline!
- from a student perspective, the test output between unittest and pytest _is_ different, but I will declare it as "a skill to work with the dissonance."
- I also think unittest output is readable enough

This PR is actually just refactoring the unittest tests

### DIFF
--- a/intro-to-tests/intro-to-pytest.md
+++ b/intro-to-tests/intro-to-pytest.md
@@ -228,11 +228,11 @@ def func_that_always_returns_true():
 ### !tests
 ```python
 import unittest
-import main as p
+from main import *
 
 class TestChallenge(unittest.TestCase):
   def test_returns_true(self):
-      self.assertTrue(p.func_that_always_returns_true())
+      self.assertTrue(func_that_always_returns_true())
 ```
 ### !end-tests
 ### !explanation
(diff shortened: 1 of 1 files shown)


CheezItMan (approved):
This looks like both the only choice we have, and a reasonable one.   This reviewer fully endorses the choice!  I like the use of the callout and the simplified testing syntax here. 

I also like that we do teach them the basics of writing a pytest!

👍

intro-to-tests/intro-to-pytest.md
```diff
@@ -192,6 +192,18 @@ def test_zero_division():
...
+### !callout-danger
+
+## This Curriculum Uses unittest, Not pytest
+This curriculum teaches pytest, a testing framework that has wide adoption in the Python community. This curriculum's projects will use pytest. However, **the code challenges within this curriculum will use unittest,** a different testing framework. This is due to a limitation in the curriculum tools.
+
+---
+
+Working with unittest, rather than pytest, will have some differences in syntax. However, conceptually, they are the exact same; the structure, error output, error messages, passed test messages, etc. will all be extremely alike.
+
+This curriculum believes in a learner's ability to smoothly work with these differences.
```
cwboden, line 204: Perhaps add one more line, "In industry, teams are often evolving to use the latest and greatest tools available. As a result, it's common for developers to learn new frameworks on the job."

cwboden (approved):

PR #43, opened 2020-12-30
iterating-over-data: ports from old core repo

tildeee:
Order is:

1. Intro to Iteration
2. Iterating over Lists
3. Iterating over Dictionaries
4. Iterating and Special Cases

### DIFF
--- a/iterating-over-data/iterating-and-special-cases.md
+++ b/iterating-over-data/iterating-and-special-cases.md
@@ -129,3 +129,116 @@ print("Blast off!")
 ```
 
 ## Check for Understanding
+<!-- Question 1 -->
+<!-- prettier-ignore-start -->
+### !challenge
+
+* type: multiple-choice
+* id: NkOZhxvAK
+* title: Using a For Loop
+
+
+##### !question
+
+True of False: Once a `for` loop is executed, it cannot be stopped early. It must complete its loop from start to finish.
+
+##### !end-question
+
+##### !options
+
+* True
+* False
+
+##### !end-options
+
+##### !answer
+
+* False
+
+##### !end-answer
+
+##### !explanation
+
+A `break` statement can be used to stop a `for` loop early.
+
+##### !end-explanation
+
+### !end-challenge
+<!-- prettier-ignore-end -->
+
+<!-- Question 2 -->
+<!-- prettier-ignore-start -->
+### !challenge
+
+* type: short-answer
+* id: 4yWezZDAY
+* title: Using a Break Statement
+
+##### !question
+
+Looking at the code below, what is the role of the `continue` statement?
+
+<!-- ```python
+for i in range(33):
+  if i % 3 != 0:
+    continue
+  print(f"{i} is divisible by 3!")
+``` -->
+
+##### !end-question
+
+##### !placeholder
+
+Your explanation here...
+
+##### !end-placeholder
+
+##### !answer
+
+/.+/
+
+##### !end-answer
+
+##### !explanation
+
+For each iteration `i` is checked to see if it is evenly divisible by 3. If it's not, then the `continue` statement moves the loop to the next iteration of `i` and checks again.
+
+##### !end-explanation
+
+### !end-challenge
+<!-- prettier-ignore-end -->
+
+<!-- Question 3 -->
+<!-- prettier-ignore-start -->
+### !challenge
+
+* type: short-answer
+* id: VJnx3ZvRt
+* title: Using a Range
+
+##### !question
+
+What is the role of the `range` function? What three arguments does it take? Are they all required?
+
+##### !end-question
+
+##### !placeholder
+
+Your answer here...
+
+##### !end-placeholder
+
+##### !answer
+
+/.+/
+
+##### !end-answer
+
+##### !explanation
+
+`Range` function iterates through a sequence of numbers. It can take three arguments: `start`, `stop`, `step`. `Start` tells it what number to begin the sequence on, `stop` tells it when to stop (not including the number), `step` tells it how many times to increase per iteration. Only `stop` is a required argument.
+
+##### !end-explanation
+
+### !end-challenge
+<!-- prettier-ignore-end -->
--- a/iterating-over-data/iterating-and-special-cases.md
+++ b/iterating-over-data/iterating-and-special-cases.md
@@ -2,9 +2,9 @@
 
 ## Learning Goals
 
-- Define the "break" keyword as a way to exit a loop
-- Define the "continue" keyword as a way to advance an iteration in a loop
-- Use the "range()" syntax combined with a for loop
+- Define the `break` keyword as a way to exit a loop
+- Define the `continue` keyword as a way to advance an iteration in a loop
+- Use the `range()` syntax combined with a for loop
 
 ## Introduction
 
@@ -24,7 +24,7 @@ By learning these different pieces of syntax, we can learn to wield them wisely 
 
 ## `break` Exits a Loop
 
-`break` is a keyword that we can put in a `for` loop. When our code comes across `break`, it will exit the `for` loop entirely.
+`break` is a keyword that can be used in a `for` loop. When the Python interpreter encounters a `break`, it will immediately exit the current loop, skipping any instructions from after the `break` until the end of the loop. We also say that it _breaks out_ of the loop.
 
 ```python
 for fruit in ["apples", "oranges", "bananas"]:
@@ -52,7 +52,7 @@ for field_item in ["grass", "grass", "more grass", "four-leaf clover", "rocks", 
 
 ## for-else
 
-Turns out we can pair a `for` loop with an `else` clause! The `else` clause will run after a `for` loop finishes **if** it did **not** encounter a `break`. The `else` is on the same level of indentation as the `for`.
+It turns out we can pair a `for` loop with an `else` clause! The `else` clause will run after a `for` loop finishes **if** it did **not** encounter a `break`. The `else` keyword must be at the same level of indentation as the `for`, and be followed by a `:`. The `else` body begins indented one level on the following line.
 
 ### !end-callout
 
@@ -71,7 +71,7 @@ else:
 
 ## `continue` Advances One Loop
 
-`continue` is another keyword we can put in a `for` loop. When our code sees `continue`, it will immediately move to the next iteration.
+`continue` is another keyword that can be used in a `for` loop. When the Python interpreter encounters a `continue`, it will immediately jump to the beginning of the current loop, skipping any instructions from after the `continue` until the end of the loop. The loop will resume execution with the next item in the iteration, if one exists. Otherwise, the loop will terminate.
 
 ```python
 lucky_charms_quantities = {
@@ -102,7 +102,7 @@ The function signature of `range` is:
 range(start, stop, step)
 ```
 
-Where `start` is the number the range should start at, `stop` is the number the range should stop at, and `step` is the amount the range should increment by. `step` is an **optional** argment, and its default value is `1`.
+Where `start` is the number at which the range should start, and `stop` is the number at which the range should stop. `stop` is exclusive, which means the `stop` number itself will _not_ be included in the range. `step` is the amount by which the range should increment. `step` is an **optional** argument, and its default value is `1`.
 
 ### Ranges in `for` Loops
 
@@ -113,10 +113,9 @@ We can pair ranges with `for` loops so that we iterate over a sequence of number
 ```python
 for number in range(0, 6):
     print(number)
-
 ```
 
-Note that it printed up to `5`, and it didn't print 6. That's because the end is _non-inclusive_, or it goes **up** to the end value.
+Note that it printed up to `5`, and it didn't print `6`. Recall that the `stop` parameter, is _exclusive_. It goes **up** to, but excludes, the end value.
 
 Another example:
 
--- a/iterating-over-data/iterating-and-special-cases.md
+++ b/iterating-over-data/iterating-and-special-cases.md
@@ -1,0 +1,132 @@
+# Iterating and Special Cases
+
+## Introduction
+
+There are more things we can do with loops!
+
+Are we able to stop a loop before it gets to the last element? Are we able to use `for` loops on other data structures?
+
+By learning these different pieces of syntax, we can learn to wield them wisely whenever it becomes relevant.
+
+## Learning Goals
+
+- Define the "break" keyword as a way to exit a loop
+- Define the "continue" keyword as a way to advance an iteration in a loop
+- Use the "range()" syntax combined with a for loop
+
+## Vocabulary and Synonyms
+
+| Vocab      | Definition                                                       | Synonyms | How to Use in a Sentence                                                                     |
+| ---------- | ---------------------------------------------------------------- | -------- | -------------------------------------------------------------------------------------------- |
+| `break`    | A keyword that will exit an entire loop                          | -        | "We can break out of the loop using `break`," "The loop exited early because it saw `break`" |
+| `continue` | A keyword that will immediately advance one interation in a loop | -        | "The loop moved on because it saw `continue`"                                                |
+| `range()`  | A Python function that creates a sequence                        | -        | "We used a `for` loop and `range()` in order to iterate through a series of numbers.         |
+
+## `break` Exits a Loop
+
+`break` is a keyword that we can put in a `for` loop. When our code comes across `break`, it will exit the `for` loop entirely.
+
+```python
+for fruit in ["apples", "oranges", "bananas"]:
+    print("I'm in the for loop! The value of fruit is:", fruit)
+    break
+    print("This message will never print, because the for loop saw break before it saw me...")
+```
+
+We can pair this with conditionals (and other logic)!
+
+```python
+print("I'm looking for a four leaf clover in the field...")
+
+for field_item in ["grass", "grass", "more grass", "four-leaf clover", "rocks", "rocks", "more rocks"]:
+    print("...")
+    if field_item == "four-leaf clover":
+        print("I found a four-leaf clover!")
+        print("My hunt is over! I can leave now")
+        break
+    print(f"I found {field_item} in the field...")
+    print("Better keep looking.")
+```
+
+### !callout-info
+
+## for-else
+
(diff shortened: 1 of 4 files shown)


iterating-over-data/iterating-and-special-cases.md
```diff
@@ -0,0 +1,132 @@
...
+
+| Vocab      | Definition                                                       | Synonyms | How to Use in a Sentence                                                                     |
+| ---------- | ---------------------------------------------------------------- | -------- | -------------------------------------------------------------------------------------------- |
+| `break`    | A keyword that will exit an entire loop                          | -        | "We can break out of the loop using `break`," "The loop exited early because it saw `break`" |
+| `continue` | A keyword that will immediately advance one interation in a loop | -        | "The loop moved on because it saw `continue`"                                                |
+| `range()`  | A Python function that creates a sequence                        | -        | "We used a `for` loop and `range()` in order to iterate through a series of numbers.         |
+
+## `break` Exits a Loop
+
+`break` is a keyword that we can put in a `for` loop. When our code comes across `break`, it will exit the `for` loop entirely.
```
anselrognlie, line 27: avoid perspective shift in description of break.
shift voicing to the interpreter hitting the break, rather than our code.
provide a bit more explicit description of what the break does.
provide a common synonymous expression.
```suggestion
`break` is a keyword that can be used in a `for` loop. When the Python interpreter encounters a `break`, it will immediately exit the current loop, skipping any instructions from after the `break` until the end of the loop. We also say that it _breaks out_ of the loop.
```

iterating-over-data/iterating-and-special-cases.md
```diff
@@ -0,0 +1,132 @@
...
+        break
+    print(f"I found {field_item} in the field...")
+    print("Better keep looking.")
+```
+
+### !callout-info
+
+## for-else
+
+Turns out we can pair a `for` loop with an `else` clause! The `else` clause will run after a `for` loop finishes **if** it did **not** encounter a `break`. The `else` is on the same level of indentation as the `for`.
```
anselrognlie, line 55: make a callback to where `else` was first introduced.
additional explicit details about the structure of the else clause
```suggestion
We have seen `else` clauses used with `if` statements, but it turns out we can also pair them with `for` loops! The `else` clause will run after a `for` loop finishes **if** it did **not** encounter a `break`. The `else` keyword must be at the same level of indentation as the `for`, and be followed by a `:`. The `else` body begins indented one level on the following line.
```

iterating-over-data/iterating-and-special-cases.md
```diff
@@ -0,0 +1,132 @@
...
+        break
+    print(f"I found {field_item} in the field...")
+    print("Better keep looking.")
+else:
+    print("I didn't find a four-leaf clover today.")
+```
+
+## `continue` Advances One Loop
+
+`continue` is another keyword we can put in a `for` loop. When our code sees `continue`, it will immediately move to the next iteration.
```
anselrognlie, line 74: avoid perspective shift in description of break.
shift voicing to the interpreter hitting the break, rather than our code.
provide a bit more explicit description of what the continue does.

```suggestion
`continue` is another keyword that can be used in a `for` loop. When the Python interpreter encounters a `continue`, it will immediately jump to the beginning of the current loop, skipping any instructions from after the `continue` until the end of the loop. The loop will resume execution with the next item in the iteration, if one exists. Otherwise, the loop will terminate.
```

iterating-over-data/iterating-and-special-cases.md
```diff
@@ -0,0 +1,132 @@
...
+- 3, 6, 9, 12, 15
+- 0, -1, -2, -3.
+
+The function signature of `range` is:
+
+```
+range(start, stop, step)
+```
+
+Where `start` is the number the range should start at, `stop` is the number the range should stop at, and `step` is the amount the range should increment by. `step` is an **optional** argment, and its default value is `1`.
```
anselrognlie, line 105: argument typo.
For this I keep going back and forth about whether I think it would be more clear or more confusing to say that stop is one beyond the number the range should stop at.  As it is, the reader might gloss over the example below and later be surprised, while rephrasing here might make them sit up and take notice, since it sounds strange, but accurate.
```suggestion
Where `start` is the number the range should start at, `stop` is the number the range should stop at, and `step` is the amount the range should increment by. `step` is an **optional** argument, and its default value is `1`.
```
or with the range option
```suggestion
Where `start` is the number the range should start at, `stop` is _one beyond_ the number the range should stop at, and `step` is the amount the range should increment by. `step` is an **optional** argument, and its default value is `1`.
```
and for ultimate pedantry, avoiding dangling prepositions
```suggestion
Where `start` is the number at which the range should start, `stop` is _one beyond_ the number at which the range should stop, and `step` is the amount by which the range should increment. `step` is an **optional** argument, and its default value is `1`.
```

iterating-over-data/iterating-and-special-cases.md
```diff
@@ -0,0 +1,132 @@
...
+### Ranges in `for` Loops
+
+We can pair ranges with `for` loops so that we iterate over a sequence of numbers. To do this, instead of a list or dictionary, we put in a range.
+
+**Look at this example:**
+
+```python
+for number in range(0, 6):
+    print(number)
+
```
anselrognlie, line 116: delete blank line
```suggestion
```

iterating-over-data/iterating-and-special-cases.md
```diff
@@ -0,0 +1,132 @@
...
+```
+
+Note that it printed up to `5`, and it didn't print 6. That's because the end is _non-inclusive_, or it goes **up** to the end value.
```
anselrognlie, line 119: use code quotes around both number, or neither.
prefer exclusive to non-inclusive
```suggestion
Note that it printed up to `5`, and it didn't print `6`. That's because the end is _exclusive_. It goes **up** to , but excludes, the end value.
```

iterating-over-data/iterating-and-special-cases.md
```diff
@@ -0,0 +1,132 @@
...
+```python
+print("Blast off in:")
+
+for count in range(10, 0, -1):
+    print(count, "...")
+
+print("Blast off!")
+```
+
+## Check for Understanding
```
spitsfire, line 132: added a few CPU questions

```suggestion
## Check for Understanding
<!-- Question 1 -->
<!-- prettier-ignore-start -->
### !challenge

* type: multiple-choice
* id: NkOZhxvAK
* title: Using a For Loop


##### !question

True of False: Once a `for` loop is executed, it cannot be stopped early. It must complete its loop from start to finish.

##### !end-question

##### !options

* True
* False

##### !end-options

##### !answer

* False

##### !end-answer

##### !explanation

A `break` statement can be used to stop a `for` loop early.

##### !end-explanation

### !end-challenge
<!-- prettier-ignore-end -->

<!-- Question 2 -->
<!-- prettier-ignore-start -->
### !challenge

* type: short-answer
* id: 4yWezZDAY
* title: Using a Break Statement

##### !question

Looking at the code below, what is the role of the `continue` statement?

<!-- ```python
for i in range(33):
  if i % 3 != 0:
    continue
  print(f"{i} is divisible by 3!")
``` -->

##### !end-question

##### !placeholder

Your explanation here...

##### !end-placeholder

##### !answer

/.+/

##### !end-answer

##### !explanation

For each iteration `i` is checked to see if it is evenly divisible by 3. If it's not, then the `continue` statement moves the loop to the next iteration of `i` and checks again.

##### !end-explanation

### !end-challenge
<!-- prettier-ignore-end -->

<!-- Question 3 -->
<!-- prettier-ignore-start -->
### !challenge

* type: short-answer
* id: VJnx3ZvRt
* title: Using a Range

##### !question

What is the role of the `range` function? What three arguments does it take? Are they all required?

##### !end-question

##### !placeholder

Your answer here...

##### !end-placeholder

##### !answer

/.+/

##### !end-answer

##### !explanation

`Range` function iterates through a sequence of numbers. It can take three arguments: `start`, `stop`, `step`. `Start` tells it what number to begin the sequence on, `stop` tells it when to stop (not including the number), `step` tells it how many times to increase per iteration. Only `stop` is a required argument.

##### !end-explanation

### !end-challenge
<!-- prettier-ignore-end -->
```

spitsfire (changes requested):
added CPU questions to iterating and special cases

iterating-over-data/iterating-and-special-cases.md
tildeee:
```suggestion
Turns out we can pair a `for` loop with an `else` clause! The `else` clause will run after a `for` loop finishes **if** it did **not** encounter a `break`. The `else` keyword must be at the same level of indentation as the `for`, and be followed by a `:`. The `else` body begins indented one level on the following line.
```

iterating-over-data/iterating-and-special-cases.md
tildeee:
```suggestion
Where `start` is the number the range should start at, `stop` is the number the range should stop at, and `step` is the amount the range should increment by. `step` is an **optional** argment, and its default value is `1`.

Where `start` is the number at which the range should start, and `stop` is the number at which the range should stop. `stop` is non-inclusive, which means the `stop` number itself will _not_ be included in the range. `step` is the amount by which the range should increment. `step` is an **optional** argument, and its default value is `1`.
```

iterating-over-data/iterating-over-dictionaries.md
```diff
@@ -0,0 +1,321 @@
...
+
+## Learning Goals
+
+- Practice using for loops to iterate through a dictionary
+
+## Introduction
+
+The overall concept for iterating over dictionaries is the same as iterating over lists. We can use `for` loops to iterate over **each key-value pair in a dictionary.**
+
+However, because dictionaries are key-value pairs, there are a few differences to pay attention to.
```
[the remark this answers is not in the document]
tildeee:
```suggestion
The overall concept for iterating over dictionaries is the same as iterating over lists. We use `for` loops to iterate over each dictionary item.

However, because dictionary items are key-value pairs, there are a few ways that we can iterate over them: **by key-value pair**, **by key alone**, or **by value alone**.
```

Gonna see how it feels to drop your very nice transition sentence.

iterating-over-data/iterating-over-dictionaries.md
```diff
@@ -0,0 +1,321 @@
...
+
+```python
+for my_key, my_value in my_dict.items():
+    print(my_key, my_value)
+```
+
+| Piece of Code              | Notes                                                                                                                                                        |
+| -------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `for`                      | `for` is a reserved keyword in Python. Python recognizes `for` as the beginning of a `for` loop.                                                             |
+| `my_key`                   | **Replace this** part with a name that represents what each key is. This local variable represents the _key_ in each key-value pair, one at a time.          |
```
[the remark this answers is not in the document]
tildeee:
```suggestion
| `my_key`                   | **Replace this** part with a name that represents what each key is. During iteration, this will be a variable that will receive the _key_ in each key-value pair, one at a time.          |
```

iterating-over-data/iterating-over-dictionaries.md
```diff
@@ -0,0 +1,321 @@
...
+| `my_value`                 | **Replace this** part with a name that represents what each value is. This local variable represents the _value_ in each key-value pair, one at a time.      |
```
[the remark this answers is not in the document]
tildeee:
```suggestion
| `my_value`                 | **Replace this** part with a name that represents what each value is. During iteration, this will be a variable which will receive the _key_ in each key-value pair, one at a time.      |
```

iterating-over-data/iterating-and-special-cases.md
anselrognlie:
```suggestion
It turns out we can pair a `for` loop with an `else` clause! The `else` clause will run after a `for` loop finishes **if** it did **not** encounter a `break`. The `else` keyword must be at the same level of indentation as the `for`, and be followed by a `:`. The `else` body begins indented one level on the following line.
```

iterating-over-data/iterating-and-special-cases.md
anselrognlie: Bringing the introduction of exclusive up to this point makes the latter mention sound redundant with a little smoothing.  I think it's fine to include here, but take a look at the suggestion for the mention below.

iterating-over-data/iterating-and-special-cases.md
anselrognlie: change for if exclusive is introduced above
```suggestion
Note that it printed up to `5`, and it didn't print `6`. Recall that the `stop` parameter, is _exclusive_. It goes **up** to, but excludes, the end value.
```

iterating-over-data/iterating-over-dictionaries.md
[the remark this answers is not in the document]
anselrognlie: My concern about this phrasing (letting During iteration decorate the whole phrase) is that it's unclear whether the variable is only accessible during the iteration (it isn't).  Maybe this???
```suggestion
| `my_key`                   | **Replace this** part with a name that represents what each key is. This will be used as a variable name that will receive the _key_ in each key-value pair during the iteration, one key at a time.          |
```

iterating-over-data/iterating-over-dictionaries.md
```diff
@@ -0,0 +1,321 @@
...
+
+### About `.items()`
+
+The most unique part about iterating over dictionaries is that we should call `.items()` next to each dictionary when starting the for loop.
+
+Python is unable to iterate over dictionaries as is; `.items()` lets us use the `for` loop with it. If we neglect to add `.items()`, we'll likely get this error:
+
+```bash
+ValueError: too many values to unpack (expected 2)
+```
```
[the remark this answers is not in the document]
tildeee: I'd like to go without including this for C15

And, I have NOT figured out how to make suggestions compatible with code blocks within them 🔥 😡 

I can't recreate your `ValueError` message; I'm going to stick with the original message 1. out of momentum 2. they're hopefully conceptually congruent enough (<- I made up that phrase) that they can be bridged, and I'd rather a student see the less-detailed one and accept the more detailed one as possible, rather than the other way around

iterating-over-data/iterating-over-dictionaries.md
```diff
@@ -0,0 +1,321 @@
...
+*
+```python
+for dessert in dessert_menu.values():
+  print(dessert)
+```
+*
+```python
+for category, dessert in dessert_menu.items():
+  print(dessert)
+```
```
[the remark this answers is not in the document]
tildeee: Good thought-- I'll keep in mind to track this question specifically, let's try C15 with it now! I'm very into teaching "what is syntactically possible" as a separate lesson besides "best solution" (obviously shouldn't teach blatant anti-patterns, but atm this is a syntax bridge I'd like students to be able to read, with less emphasis on evaluation)

iterating-over-data/iterating-over-lists.md
```diff
@@ -0,0 +1,209 @@
...
+The variable `my_element` gets assigned and re-assigned through a `for` loop.
+
+When a `for` loop starts, `my_element` gets assigned the value of the first element in the list. Then value of `my_element` changes with each following iteration.
+
+1. In the first iteration, the value of `my_element` will be the first element in the list.
+2. In the second iteration, the value of `my_element` will be the second element in the list.
+3. In the third, it will be the third element...
+4. ... etc., until there are no more elements in the list.
+
+We should always change `my_element` to be a variable name appropriate to the context of the list. What does each element in the list represent?
```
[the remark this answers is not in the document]
tildeee: Taking your suggestion and shortening it

```suggestion
We should always change `my_element` to be a variable name appropriate to the context of the list. Think about what each element in the list represents. If we are iterating over a list of books, we might change `my_element` to `book`. For a list of rooms, the name `room` might be more appropriate.

The name we choose doesn't affect the underlying iteration value in any way, but it can help us remember what data we're processing in the loop body.
```

iterating-over-data/iterating-over-lists.md
```diff
@@ -0,0 +1,209 @@
...
+| 3           | 2                | `48.50`          | `53.3985`                                     |
+| 4           | 3                | `18.50`          | `20.3685`                                     |
+
+## Variable Scope
+
+Inside `for` loops, we can access and use any variables that are within scope.
+
+A common pattern we'll use is to initialize a variable before a `for` loop, and modify it within a `for` loop.
+
+Read through this code, and answer these questions:
```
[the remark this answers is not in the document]
tildeee: Perfect!

iterating-over-data/iterating-over-dictionaries.md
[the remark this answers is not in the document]
anselrognlie: Weird... Out of curiosity, what code were you running?  I did 
```python
my_dict = { 'a': 0, 'b': 1}
for k, v in my_dict:
    print(k, v)
```

And to stick with the shorter content, I'd still make some suggestions, specifically targeted at focusing the material on the iteration described here being the key-value pairs. For instance, Python _is_ able to iterate over dictionaries without .items(), it just iterates over keys.  Maybe something like this?

==================

The most unique part about iterating over dictionary key-value pairs is that we must use `.items()` with the dictionary when starting the `for` loop. Notice that it appears right after the dictionary variable, with no space in between.

This syntax is required because Python doesn't iterate over the _items_ of a dictionary by default. `.items()` informs the dictionary that we'd like to use the key-value pairs in the `for` loop instead of the dictionary's default behavior. If we neglect to add `.items()`, we'll likely get this error:

```bash
ValueError: too many values to unpack (expected 2)
```

iterating-over-data/iterating-over-lists.md
```diff
@@ -0,0 +1,209 @@
...
+- ... accessed or modified inside the `for` loop? How?
+- ... accessed or modified after the `for` loop? How?
+
+```python
+prices = [18.99, 56.00, 48.50, 18.50]
+taxed_prices = []
+
+for price in prices:
+    taxed_price = price * 1.101
+    taxed_prices.append(taxed_price)
```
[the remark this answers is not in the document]
jmaddox19: Yes! Please let's do this!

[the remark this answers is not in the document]
jmaddox19: I guess I could also see us having the opinion that we want to start by teaching them one way to do things and letting them learn about more optimal ways of doing things later on...?

PR #48, opened 2021-01-07
math-primer: adds content

tildeee:


fundamentals-recap/math-primer.md
```diff
@@ -0,0 +1,170 @@
...
+
+### !callout-info
+
+## Origin of Base 10
+
+"Legend says" that most base-ten systems possibly originate from counting on ten fingers.
+
+### !end-callout
+
+## Using _`n`_ as a Variable in Math
```
anselrognlie, line 38: To me, this section really blurs the line between variables as a mathematical concept, and variables as a programming concept. For the brevity of the material, that may be fine, but if this is meant to help students be more comfortable with mathematical terminology and expressions (which often come up in computing), then keeping thing focused on variables in equations, and looking at simple algebraic operations might be more useful.

What really is the expected level of mathematics for incoming students? I do occasionally encounter students who don't seem very comfortable expressing mathematical relationships between quantities with equations. Having a good grounding in math is very useful for any programmer. While the classroom curriculum is too short for a full math primer, it would be great to provide some links to external sources that students could use to fill in the gaps, and if there are core expectations, I'd like to see the Ada curriculum include that information, at least in a reference form.

fundamentals-recap/math-primer.md
```diff
@@ -0,0 +1,170 @@
...
+
+When we describe problems, we might want to say, "any number of things."
+
+For example, we might be building an email webapp, and we want to say, "An email can have any number of recipients... It can have 1 recipient, 3, 13, or 50 recipients."
+
+We wouldn't want to say "An email can have infinite recipients," because that just may not be true. However, when we want to discuss numbers of things, we may start to use variables as we talk.
+
+We can imagine a variable _`n`_ as the name of a variable that holds any number.
```
anselrognlie, line 46: suggest avoiding invocation of infinity
```suggestion
It would be annoying if we always had to say that an email might have 1, 3, 13, etc., recipients. Instead, we can use a variable to talk about this unknown number.

We can imagine a variable _`n`_ as the name of a variable that holds any number. Then we could say that the email has _`n`_ recipients.
```

anselrognlie (approved):
I'd really like to see this grow over time. Even if it's not explicitly covered during classwork, having a body of content that says 'this is the math we'll expect students to be able to discuss and apply to their work' would be helpful for individuals without a math background.

fundamentals-recap/math-primer.md
```diff
@@ -0,0 +1,170 @@
+# Math Primer
+
+## Goal
+
+Problem-solving in programming is almost always centered on _data_. Often times, it's useful to be aware of the following math concepts. Knowing the definition of these concepts can become tools you remember to use in the future.
```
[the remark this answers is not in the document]
tildeee: Thanks, I particularly like this smoothing! This might be a good "ideal" style for me to aim for, I appreciate it!

fundamentals-recap/math-primer.md
```diff
@@ -0,0 +1,170 @@
...
+Working with collections of numbers:
+
+1. Maximum and Minimum
+1. Sorting
+1. Average or Mean
+1. Floor and Ceiling Functions
+
+## Decimal Number System
+
+The _decimal number system_ or _base-ten number system_ is the numbering system most humans use in everyday life.
```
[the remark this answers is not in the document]
tildeee: I definitely remember writing this and being like "I'm going to regret writing this, humans are all very different" 😆

fundamentals-recap/math-primer.md
tildeee: Thank you for that! (I have a bad habit of using infinity!)

fundamentals-recap/math-primer.md
tildeee: Good point.

I'll keep thinking on this; I'm working on determining the right data for what math is expected at Ada. 😄  I'm mostly trying to convert [this document](https://github.com/Ada-Developers-Academy/textbook-curriculum/blob/master/04-cs-fundamentals/classroom/Essential%20Mathematics.md) for ease of transition from C14-C15!

fundamentals-recap/math-primer.md
```diff
@@ -0,0 +1,175 @@
...
+
+The goal of this document is to be a resource. We recommend spending your energy on this document reading through it and taking notes on it once, and then reviewing briefly every few months.
+
+This document covers the following:
+
+Thinking about numbers:
+
+1. Decimal Number System
+1. Using _n_ as a Variable in Math
+1. Coordinate Planes
```
audreyandoy, line 15: Will integers be introduced with coordinate planes or as a separate item?

fundamentals-recap/math-primer.md
```diff
@@ -0,0 +1,175 @@
...
+
+Working with collections of numbers:
+
+1. Maximum and Minimum
+1. Sorting
+1. Average or Mean
+1. Floor and Ceiling Functions
+
```
audreyandoy, line 23: Modulo would be a good addition in the future as it's often confused with division to newer students.

audreyandoy (approved):
This looks great!! I'm looking forward to seeing how this grows in the future.

PR #50, opened 2021-01-11
Practice iterating over data

spitsfire:
One thing to look over: Question 7. We show an example of a `for-else` loop using print statements and a break case; however, this can also be done with a `return` statement, which is how my test is set up. Seems more likely they will be using `return` statements instead of print and break cases.

### DIFF
--- a/iterating-over-data/practice-iterating-over-data.md
+++ b/iterating-over-data/practice-iterating-over-data.md
@@ -373,6 +373,7 @@ def search_dictionary(dict,check):
 
 ```
 
+##### !end-explanation
 
 ### !end-challenge
 <!-- prettier-ignore-end -->
@@ -458,7 +459,7 @@ def search_dictionary(dict,check):
   return False
 
 ```
-
+##### !end-explanation
 
 ### !end-challenge
 <!-- prettier-ignore-end -->
--- a/iterating-over-data/practice-iterating-over-data.md
+++ b/iterating-over-data/practice-iterating-over-data.md
@@ -9,93 +9,6 @@ Complete all questions below.
 <!-- Question 1 -->
 <!-- prettier-ignore-start -->
 ### !challenge
-* type: multiple-choice
-* id: EJew-RW0t
-* title: Iterating Over Data
-##### !question
-
-Given a function called `stopwatch` that takes in the argument `22`, which of the following loops would start at `1` and increment all the way up to *and include* `22`?
-
-##### !end-question
-##### !options
-
-* `for i in 22:`
-* `for i in range(22):`
-* `for i in range(1, 22):`
-* `for i in range(1, 23):`
-
-##### !end-options
-##### !answer
-
-* `for i in range(1, 23):`
-
-##### !end-answer
-##### !explanation
-
-Python's `range` has an *exclusive nature*, meaning that it does not include the last integer in its range, ie the `stop` parameter. That's why we must use `23` instead of just `22`.
-
-##### !end-explanation
-### !end-challenge
-<!-- prettier-ignore-end -->
-
-<!-- Question 2 -->
-<!-- prettier-ignore-start -->
-### !challenge
-* type: code-snippet
-* language: python3.6
-* id: EkdvU6Z0K
-* title: Iterating Over Data
-##### !question
-
-Create a function named `sum_list` that takes a parameter `list` (a list of integers). Iterate through the list and sum up each integer. Then return the total.
-
-##### !end-question
-##### !placeholder
-
-```python
-
-def sum_list(list):
-    # Note: the word "list" may be highlighted because the Learn platform believes it's reserved
-    # This should not affect our code
-    pass
-
-```
-
-##### !end-placeholder
-##### !tests
-```python
-
-import unittest
-from main import *
-
-class TestPython1(unittest.TestCase):
-    def test_sum_list(self):
-        self.assertEqual(sum_list([3,5,7,9]),24)
-        self.assertEqual(sum_list([0,-3,6,6]),9)
-
-```
-##### !end-tests
-##### !explanation
-
-An example of a working implementation:
-
-``` python
-
-def sum_list(list):
-  sum = 0
-  for item in list:
-    sum += item
-  return sum
-
-```
-
-##### !end-explanation
-### !end-challenge
-<!-- prettier-ignore-end -->
-
-<!-- Question 3 -->
-<!-- prettier-ignore-start -->
-### !challenge
 * type: checkbox
 * id: N1iCy1fCF
 * title: Iterating Over Data
@@ -166,7 +79,7 @@ This question purposefully uses unreadable variable names. Replace the variable 
 ### !end-challenge
 <!-- prettier-ignore-end -->
 
-<!-- Question 4 -->
+<!-- Question 2 -->
 <!-- prettier-ignore-start -->
 ### !challenge
 * type: checkbox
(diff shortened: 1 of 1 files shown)


iterating-over-data/practice-iterating-over-data.md
```diff
@@ -0,0 +1,464 @@
...
+Which loop(s) *can* be used to print out each key?
+
+##### !end-question
+
+##### !options
+
+* `for x in my_profile:`
+* `for x in my_profile.keys()`
+* `for x in my_profile.values()`
+* `for x, y in my_profile.items()`
```
spitsfire, line 155: these variables are purposeful. I wanted the student to focus on answering these mc questions by the dictionary methods, NOT clues from good descriptive variable names

iterating-over-data/practice-iterating-over-data.md
```diff
@@ -0,0 +1,465 @@
...
+Given a function called `stopwatch` that takes in the argument `22`, which of the following loops would  correctly start at `1` and increment all the way up to *and include* `22`?
+
+##### !end-question
+
+##### !options
+
+* `for i in 22:`
+* `for i in range(22)`
+* `for i in range(1,22)`
+* `for i in range(1,23)`
```
audreyandoy, line 28:
```suggestion
* `for i in range(22):`
* `for i in range(1,22):`
* `for i in range(1,23):`
```

Adding in consistency in syntax

audreyandoy (approved):

iterating-over-data/practice-iterating-over-data.md
```diff
@@ -0,0 +1,465 @@
...
+<!-- prettier-ignore-start -->
+### !challenge
+
+* type: multiple-choice
+* id: EJew-RW0t
+* title: Which Range is Correct?
+
+##### !question
+
+Given a function called `stopwatch` that takes in the argument `22`, which of the following loops would  correctly start at `1` and increment all the way up to *and include* `22`?
```
tildeee, line 19: Dropped accidental double-space, remove "correctly"-- the question is inherently asking for the correct solution

```suggestion
Given a function called `stopwatch` that takes in the argument `22`, which of the following loops would start at `1` and increment all the way up to *and include* `22`?
```

iterating-over-data/practice-iterating-over-data.md
```diff
@@ -0,0 +1,465 @@
...
+* `for i in 22:`
+* `for i in range(22)`
+* `for i in range(1,22)`
+* `for i in range(1,23)`
+
+##### !end-options
+
+##### !answer
+
+`for i in range(1,23)`
```
tildeee, line 34: Adding bullet point for format consistency

```suggestion
* `for i in range(1,23):`
```

@audreyandoy FYI, when changing options, the answer should also be updated!

iterating-over-data/practice-iterating-over-data.md
```diff
@@ -0,0 +1,465 @@
...
+##### !tests
+```python
+
+import unittest
+from main import *
+
+class TestPython1(unittest.TestCase):
+def test_sum_list(self):
+  self.assertEqual(sum_list([3,5,7,9]),24)
+  self.assertEqual(sum_list([0,-3,6,6]),9)
```
tildeee, line 93: Incorrect indentation breaks this question

```suggestion
class TestPython1(unittest.TestCase):
    def test_sum_list(self):
        self.assertEqual(sum_list([3,5,7,9]),24)
        self.assertEqual(sum_list([0,-3,6,6]),9)
```

iterating-over-data/practice-iterating-over-data.md
tildeee, line 19: We technically don't teach `range()`, but even if we did, this question assesses knowledge about `range()`, which seems a little out of place here? I'd be open for pushback because sometimes reinforcement like this is good, but the fact that it's both not-taught and out of place makes me want to remove it.

iterating-over-data/practice-iterating-over-data.md
tildeee: Nitpick: Is there a reason why we aren't doing a space between `1,` and `2X`?

```suggestion
* `for i in range(22):`
* `for i in range(1, 22):`
* `for i in range(1, 23):`
```

iterating-over-data/practice-iterating-over-data.md
```diff
@@ -0,0 +1,465 @@
...
+Create a function named `sum_list` that takes a parameter `list` (a list of integers). Iterate through the list and sum up each integer. Then return the total.
+
+##### !end-question
+
+##### !placeholder
+
+```python
+
+def sum_list(list):
+  pass
```
tildeee, line 78: Learn does syntax highlighting with the keyword `list`, which could be really jarring (in the past, Ada students have been really confused at what gets highlighted and why it does, taking up much conversation)

I suggest either changing the var name `list`, or adding a quick explanation.

In another lesson, I've added this note.

```suggestion
def sum_list(list):
    # Note: the word "id" may be highlighted because the Learn platform believes it's reserved
    # This should not affect our code
    pass
```

Also, we should conform to 4 space indents

iterating-over-data/practice-iterating-over-data.md
```diff
@@ -0,0 +1,465 @@
...
+* `for x, y in my_profile.items()`
+
+
+##### !end-options
+
+##### !answer
+
+* `for x in my_profile:`
+* `for x in my_profile.keys()`
+* `for x, y in my_profile.items()`
```
tildeee, line 164: Make colons or no colons consistent

```suggestion
* `for x in my_profile:`
* `for x in my_profile.keys():`
* `for x in my_profile.values():`
* `for x, y in my_profile.items():`


##### !end-options

##### !answer

* `for x in my_profile:`
* `for x in my_profile.keys():`
* `for x, y in my_profile.items():`
```

iterating-over-data/practice-iterating-over-data.md
```diff
@@ -0,0 +1,465 @@
...
+
+
+##### !end-options
+
+##### !answer
+
+* `for x in my_profile:`
+* `for x in my_profile.values()`
+* `for x, y in my_profile.items()`
```
tildeee, line 222: This answer should have all of them, technically? I have a note similar to this below

iterating-over-data/practice-iterating-over-data.md
```diff
@@ -0,0 +1,465 @@
...
+For example, given: `{"icecream_flavor": "chocolate", "lucky_nums": [7, 18, 34], has_pets: True}` and `has_pets`, return `True`.
+
+##### !end-question
+
+##### !placeholder
+
+```python
+
+def search_dictionary(dict,check):
+  pass
```
tildeee, line 323: Same thing as before: Learn ends up highlighting `dict` so I recommend using a diff var name or adding this note.

Also, add a space between `dict,` and `check`

```suggestion
def search_dictionary(dict, check):
    # Note: the word "id" may be highlighted because the Learn platform believes it's reserved
    # This should not affect our code
    pass
```

iterating-over-data/practice-iterating-over-data.md
```diff
@@ -0,0 +1,465 @@
...
+### !challenge
+
+* type: code-snippet
+* language: python3.6
+* id: NkbCNAWCF
+* title: Match the Output
+
+##### !question
+
+Given a function named `search_dictionary` that takes two parameters: `dict` (a dictionary) and `check` (a string), check if the key `check` is found in `dict`, and then return the value at that key.
```
tildeee, line 312: `check` isn't my favorite var name, could we do something like `needle`? (The placeholder code should be changed if this is the case)

iterating-over-data/practice-iterating-over-data.md
```diff
@@ -0,0 +1,465 @@
...
+
+For example, given: `{"icecream_flavor": "chocolate", "lucky_nums": [7, 18, 34], has_pets: True}` and `has_pets`, return `True`.
```
tildeee, line 314: For this specific example, I skimmed through it quickly, and almost thought the question was "return `True` if the key was found" -- which isn't the case, it just happens that `has_pets` is `True`

For clarity, I'd like to pick another key-value pair in this example?

iterating-over-data/practice-iterating-over-data.md
```diff
@@ -0,0 +1,465 @@
...
+```python
+
+import unittest
+from main import *
+
+dict = {"icecream_flavor": "chocolate", "lucky_nums": [7, 18, 34], has_pets: True}
+class TestPython1(unittest.TestCase):
+def test_one(self):
+  self.assertEqual(search_dictionary(dict, "has_dogs"))
+  self.assertEqual(search_dictionary(dict, "icecream_flavor"),"chocolate")
```
tildeee, line 339: The broken indentation makes this test not work

Also, what is the `self.assertEqual(search_dictionary(dict, "has_dogs"))` test?

```suggestion
class TestPython1(unittest.TestCase):
    def test_one(self):
        self.assertEqual(search_dictionary(dict, "has_dogs"))
        self.assertEqual(search_dictionary(dict, "icecream_flavor"),"chocolate")
```

iterating-over-data/practice-iterating-over-data.md
```diff
@@ -0,0 +1,465 @@
...
+```python
+
+import unittest
+from main import *
+
+dict = {"icecream_flavor": "chocolate", "lucky_nums": [7, 18, 34], has_pets: True}
+class TestPython1(unittest.TestCase):
+def test_one(self):
+  self.assertEqual(search_dictionary(dict, "has_dogs"),False)
+  self.assertEqual(search_dictionary(dict, "icecream_flavor"),"chocolate")
```
tildeee, line 421: Same as above, the broken indentation breaks the tests and they can't run

tildeee (changes requested):
Overall, the indentation problems break a lot of the code, and our placeholders should conform to four tab spaces

There are a few questions that I think aren't as effective as they could be -- the ones about picking loops that can access keys/values is tough bc dictionaries can do that in a few different ways, and in ways that Ada students will have questions on.

Submitting a formal PR review, but I'll take the lead from here!

iterating-over-data/practice-iterating-over-data.md
tildeee, line 425: Why is `dict` defined outside of the unit test? It's totally appropriate to make it a variable local to the test itself

```suggestion
class TestPython1(unittest.TestCase):
    def test_one(self):
        dict = {"icecream_flavor": "chocolate", "lucky_nums": [7, 18, 34], has_pets: True}
        self.assertEqual(search_dictionary(dict, "has_dogs"),False)
        self.assertEqual(search_dictionary(dict, "icecream_flavor"),"chocolate")
```

iterating-over-data/practice-iterating-over-data.md
tildeee: `has_pets` needs to be `"has_pets"`

iterating-over-data/practice-iterating-over-data.md
tildeee: I was wrong, apparently we DO teach range. Huh. LOOL

tildeee (approved):

PR #61, opened 2021-02-09
variables-are-references: content

tildeee:


functions-continued/variables-are-references.md
```diff
@@ -0,0 +1,648 @@
...
+[After modification] value of mangoes ['hearts', 'stars', 'horseshoes', 'clovers']
+```
+
+_Even after modification_, the mutable object's object ID remains the same.
+
+## Multiple References to Mutable Objects
+
+As mentioned earlier, variables are references to objects in memory.
+
+Two different variables can end up referring to _the same_ object in memory.
```
anselrognlie, line 289: I discussed this situation with immutable objects as well, so now this tone sounds strange.  Add some transition text to recognize the similarity to the previous discussion.

```suggestion
As with immutable objects, we can end up with two different variables referring to _the same_ object in memory. But with mutable objects, the situation is a little more interesting.
```

functions-continued/variables-are-references.md
```diff
@@ -0,0 +1,648 @@
...
+```python
+berries = ["hearts", "stars", "horseshoes"]
+
+melons = berries
+```
+
+We can describe the relationship between `melons` and `berries` in any of these following ways:
+
+- The value of `melons` is whatever the value of `berries` is
+- The variable `melons` refers to the object that `berries` refers
```
anselrognlie, line 302:
```suggestion
- The variable `melons` refers to the object to which `berries` refers
```

functions-continued/variables-are-references.md
```diff
@@ -0,0 +1,648 @@
...
+- `melons` points to `berries`
```
anselrognlie, line 303: I don't like this one, since it gives a sense of primacy to berries. melons doesn't point to berries, both melons and berries point to the same object (which is the next bullet).  So I would cut this one.

```suggestion
```

functions-continued/variables-are-references.md
```diff
@@ -0,0 +1,648 @@
...
+id of berries: 4510649216
+id of melons: 4510649216
+id of loquats: 4510649216
+```
+
+### !callout-success
+
+## Pro-Tip: Experiment!
+
+There are an infinite number of ways to assign and reassign and check this logic. Experiment in your own Python files!
```
anselrognlie, line 371: Is it infinite? Really? 😉

```suggestion
There are a huge number of ways to assign and reassign and check this logic. Experiment in your own Python files!
```

functions-continued/variables-are-references.md
```diff
@@ -0,0 +1,648 @@
...
+After modifying berries
+id of berries: 4369938880
+id of melons: 4369938880
+value of berries: ['hearts', 'stars', 'horseshoes', 'clovers']
+value of melons: ['hearts', 'stars', 'horseshoes', 'clovers']
+```
+
+Re-read through the code slowly, and take notes about how the modification to `berries` affects the value that `melons` points to.
+
+`berries` and `melons` both point to the same object. We modify the object with object ID `4369938880` with `berries.append("clovers")`. We see that the value of `melons` (which points to the object `4369938880`) is also modified.
```
anselrognlie, line 420: Add also to the parenthetical for reinforcement, and then rephrase the following 'also' so it doesn't sound repetitive.

```suggestion
`berries` and `melons` both point to the same object. We modify the object with object ID `4369938880` with `berries.append("clovers")`. We see that the value of `melons` (which also points to the object `4369938880`) is modified as well.
```

functions-continued/variables-are-references.md
```diff
@@ -0,0 +1,648 @@
...
+
+## Applied in Function Arguments
+
+When we pass in values into a function call, we should have awareness about if we are passing a mutable object or immutable object.
```
anselrognlie, line 424: a few small changes
```suggestion
When we pass values into a function call, we should be aware of whether we are passing a mutable or an immutable object.
```

functions-continued/variables-are-references.md
```diff
@@ -0,0 +1,648 @@
...
+1. `berries` points to a mutable object
+1. When we call `add_clovers`
+   - We pass in `berries`
+   - `charms` points to the same mutable object that `berries` does
+1. The object that `charms` points to is modified with `charms.append()`
+1. Because `charms` was modified, so was `berries`
+
+### Immutable Objects Passed Into Functions
+
+When we pass an immutable object into a function, that function cannot modify the object passed in. Any "modifications" are likely reassignment.
```
anselrognlie, line 478: Suggest emphasis on immutable since it's easy to lose track of which we're talking about.
Set up the following example.

```suggestion
When we pass an _immutable_ object into a function, that function cannot modify the object passed in. Any "modifications" are likely reassignment.

Recall that reassignment replaces a variable's reference with a reference to the new object. Since parameters are local variables, only the local reference is replaced. The external variable is left unchanged, as a brief example will show.
```

functions-continued/variables-are-references.md
```diff
@@ -0,0 +1,648 @@
...
+1. `jackfruit` points to an immutable object
+1. When we call `add_lucky_bonus`
+   - We pass in `jackfruit`
+   - `score` initially points to the same object as `jackfruit`
+   - After `score += 777`, `score` points to a different immutable object
+1. `jackfruit` continues to point to the same immutable object
+
+## Extending the Concept: Nested Data Structures
+
+Lists and dictionaries are not only mutable data types, but they're also _containers_. As containers, they contain either mutable or immutable data types itself!
```
anselrognlie, line 526:
```suggestion
Lists and dictionaries are not only mutable data types, but they're also _containers_. As containers, they contain either mutable or immutable data types themselves!
```

functions-continued/variables-are-references.md
```diff
@@ -0,0 +1,648 @@
...
+print("value of peaches:", peaches)
+```
+
+```
+value of peaches: [False, {'hello': 'modified value'}, False]
+```
+
+`pears` and `peaches[1]` refer to the same dictionary object.
+
+Giving a new value to a key in a dictionary counts as modification. Modifying `pear` affects the item at `peaches[1]`.
```
anselrognlie, line 599: "counts as modification" makes it sound like there are a lot of arcane rules about what things are modification, and what aren't. An intuitive understanding will probably be good enough going forward, but even if we do need further explanations, I think this has the chance of putting up a mental block before we get the chance to present the discussion.
pear -> pears

```suggestion
We modified the dictionary referred to by `pears` by storing a new value for the `"hello"` key. Modifying `pears` also affects the item at `peaches[1]`.
```

functions-continued/variables-are-references.md
```diff
@@ -0,0 +1,648 @@
...
+
+References and immutability should primarily affect our programming with awareness.
+
+Without awareness, modifying a mutable object risks unintentionally affecting other references.
+
+As general patterns to takeaway:
+
+1. Stay aware of the difference between modifying and reassigning
+1. While programming, create diagrams to help understand references between variables and objects
+1. It's more reliable to _return_ any important values from a function, rather than modify a value
```
anselrognlie, line 627: Provide justification for why we should return results rather than modifying inputs.

```suggestion
1. It's usually more desirable to _return_ important results from a function, rather than modifying a value that was passed in
    1. Callers of our functions usually don't expect the values they pass in to be modified unless there is clear documentation that your function will do so
```

functions-continued/variables-are-references.md
```diff
@@ -0,0 +1,648 @@
...
+
+### !callout-success
+
+## Foreshadowing: OOP
+
+When discussing object-oriented programming, we'll write user-defined classes. User-defined classes are mutable data types by default.
+
+### !end-callout
+
+## Best Pattern: Reassign Often, Return Often
```
anselrognlie, line 617: How about this instead? I think it emphasizes the importance of using return values, which seems like the most important take away from this.
```suggestion
## Prefer Returning Results Over Modifying Inputs
```

functions-continued/variables-are-references.md
```diff
@@ -0,0 +1,648 @@
...
+
+References and immutability should primarily affect our programming with awareness.
```
anselrognlie, line 619: I'm not sure what this was getting at, but maybe something like this?

```suggestion
We must remain aware of how references and mutable objects interact within our functions.
```

anselrognlie (changes requested):
I think this lesson got a bit into the weeds on object IDs.  Some discussion of that is definitely useful for giving a mechanism for how references work. I cut and replaced a bunch of it, since our initial discussion about leaving it as is with a disclaimer felt awkward as I read through more of this.

Happy to discuss further!

functions-continued/variables-are-references.md
```diff
@@ -0,0 +1,648 @@
+# Variables Are References
```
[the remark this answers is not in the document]
tildeee: Because yaml diffs are horrible to resolve, I typically prefer changing the config on `main` as part of the merge->update config->publish process

functions-continued/variables-are-references.md
```diff
@@ -0,0 +1,648 @@
...
+
+## Vocabulary and Synonyms
+
+| Vocab           | Definition                                                                                                                                                                                                               | Synonyms | How to Use in a Sentence |
+| --------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ | -------- | ------------------------ |
+| Object Identity | An object's unique identity that never changes once it's been created. Usually a number. Similar to an object's address in memory. Find this with `id(obj)`. In a comparison that uses `is`, Python compares object IDs. | Object's address in memory | "I knew those two objects were different because they have different object IDs," "Maybe that variable's value is changing because two variables point to the same object with the same object ID"
```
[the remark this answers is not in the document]
tildeee: In this specific case, I think Python really means [object identity](https://docs.python.org/3.7/reference/datamodel.html) (quote to pull: "the id() function returns an integer representing its identity.")

Tbh, I don't think I know enough about what you mean between Object Identifier and Object ID, I'll message you about it unless you get to this comment first?

functions-continued/variables-are-references.md
```diff
@@ -0,0 +1,648 @@
...
+
+While a program is running, every time an object is created, the object gets put into computer memory and is given an identity. We can think of this ID as the object's address in memory.
+
+We can see an object's ID by using the `id()` function.
+
+```python
+print('The ID of the string object "Hello, World!" is', id("Hello, World!"))
+```
+
+The `is` operator compares object IDs. **_An object's identity never changes once it has been created_**.
```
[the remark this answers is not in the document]
tildeee: What do you think about...

```suggestion
The `is` operator compares object IDs. **_An object's identity never changes within its lifetime_**.
```

The trick here is that I don't particularly think it's useful to define "lifetime" here.... but I the point comes across accurately and deeply enough without that definition...?

functions-continued/variables-are-references.md
```diff
@@ -0,0 +1,648 @@
...
+hello_world = "Hello, World!"
+current = 33
+is_enabled = True
+winners = [81, 231, 12]
+```
+
+- `hello_world` is a variable that refers to the object `"Hello, World!"` in memory
+- `current` is a variable that refers to the object `33` in memory
+- `is_enabled` is a variable that refers to the object `True` in memory
+- `winners` is a variable that refers to the object `[81, 231, 12]` in memory
```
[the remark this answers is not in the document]
tildeee: I'd love to make sure that I understand this well enough...

Does this wording still hold true (x out the "copy of" part)?

```suggestion
- `hello_world` is a variable that refers to the object `"Hello, World!"` in memory. It holds the object ID of `"Hello, World!"`.
- `current` is a variable that refers to the object `33` in memory. It holds the object ID of `33`.
- `is_enabled` is a variable that refers to the object `True` in memory. It holds the object ID of `True`.
- `winners` is a variable that refers to the object `[81, 231, 12]` in memory. It holds the object ID of `[81, 231, 12]`.
```

functions-continued/variables-are-references.md
```diff
@@ -0,0 +1,648 @@
...
+
+## An Analogy: Variables and Valet Parking
+
+Variables and their values can be described like valet parking services.
+
+Stevie drives their blue car and needs to park at a hotel. They are gifted _free_ valet service (so fancy). The attendant takes the car and parks it in the parking spot numbered 151. The attendant gives a ticket to Stevie and says, "when you want your car back, turn in this special ticket, and your car will come out."
+
+When Stevie wants their car back, they need to show their special ticket.
```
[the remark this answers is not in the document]
tildeee: Yeah!

Let's go with car valets for now, I like how you're musing on this analogy though. The "Stevie's friend" extension is really exciting IMO

functions-continued/variables-are-references.md
tildeee: This is an extra good note for me to think about, thanks!

functions-continued/variables-are-references.md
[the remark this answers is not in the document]
anselrognlie: Good to know.

functions-continued/variables-are-references.md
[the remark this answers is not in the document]
anselrognlie: Yeah, I've read that too. I guess my issue is that really, in the larger world, ID can refer to identity or identification, but that's not how I think of it.  I'll bring this up with my therapist. 😄

functions-continued/variables-are-references.md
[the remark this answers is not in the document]
anselrognlie: That's perfect!

functions-continued/variables-are-references.md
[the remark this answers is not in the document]
anselrognlie: It's still true. My intent in having the 'copy of' is to imply that this may not be the sole variable holding a reference to the object.  TBH, even saying that the variable holds the object ID is a little wishy washy, but I wanted to give a sense of what it means 'to refer' to an object.  i.e. the variable doesn't hold the object directly, but holds something that can be used to get to the object.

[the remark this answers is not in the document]
tildeee: What if...

```suggestion
- `hello_world` is a variable that refers to the object `"Hello, World!"` in memory. It holds something that can be used to get the object directly: a copy of the object ID of `"Hello, World!"`.
- `current` is a variable that refers to the object `33` in memory. It holds a copy of the object ID of `33`.
- `is_enabled` is a variable that refers to the object `True` in memory. It holds a copy of the object ID of `True`.
- `winners` is a variable that refers to the object `[81, 231, 12]` in memory. It holds a copy of the object ID of `[81, 231, 12]`.
```

functions-continued/variables-are-references.md
```diff
@@ -0,0 +1,648 @@
...
+
+- `list`
+- `dict`
+
+Examples of mutation and modification include:
+
+- Appending or removing an item to a list
+- Adding or removing a key-value pair to a dictionary
+
+What does this _mean_ and how does it impact programming? One of the best ways to witness the differences is to learn how a computer manages them in memory, discussed below.
```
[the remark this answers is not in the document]
tildeee: I think this is a good section, but imo, I think the second example can be explained with piggy-backing off the first example? What do you think?

Also, this rewording focuses on presenting the examples as not necessarily tricky, but some examples and a callout.

```suggestion
Examples of operations that do _not_ mutate or modify include:

- Functions that take in a variable and return some new data, such as `type(hello_world)`
- Functions that do a command, but do not change the value, such as `print(hello_world)`
- Adding two numbers, resulting in a third number, such as `sum = 1 + 2`
- Incrementing a number stored in a variable, like `num += 1`

### !callout-info

## Adding Integers Doesn't Modify

When considering `sum = 1 + 2`, it may seem like we are modifying either the 1 or the 2 to turn it into a 3, which gets assigned to the variable `sum`. In reality, summing 1 and 2 changes neither 1 nor 2. Both retain their 1-ness and 2-ness, but they give us a new value, 3! Then `sum` gets a reference to 3, not the value 3 itself.

<br>

In the case `sum += 1`, by default, this line is evaluated as `sum = sum + 1`. Therefore, it works like the situation above! 1 retains its 1-ness and `sum` gets re-assigned to the new integer value.

### !end-callout

What does this _mean_ and how does it impact programming? One of the best ways to witness the differences is to learn how a computer manages them in memory, discussed below.
```

tildeee:
I'm sorry! I made this PR horrible on accident.

There are two outstanding comments on this PR that are worthy of discussion:

https://github.com/Ada-Developers-Academy/core-unit-1/pull/61/files#r586820588
https://github.com/Ada-Developers-Academy/core-unit-1/pull/61/files#r586905088

functions-continued/variables-are-references.md
```diff
@@ -0,0 +1,609 @@
...
+
+While a program is running, every time an object is created, the object gets put into computer memory and is given an identity. We can think of this ID as the object's address in memory.
+
+We can see an object's ID by using the `id()` function.
+
+```python
+print('The ID of the string object "Hello, World!" is', id("Hello, World!"))
+```
+
+<<<<<<< HEAD
```
anselrognlie, line 49: merge conflict?

functions-continued/variables-are-references.md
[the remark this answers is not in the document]
anselrognlie: I'm starting to think this is trying to be very precise while still being imprecise.  Maybe just drop this whole line of thinking and just go with what you had originally.  Even with all the careful wording around object ID, of course, what the variables _really_ hold is a pointer to some PyObject "derived" type (which has the reference count, and the associated data-or pointer to data for more complicated types).

My main concern is whether students will have an intuitive sense of what 'refers to' means.  This would be  great spot for some visuals.  Chris' lists and references PR goes right for the jugular with memory addresses and diagrams.  Or maybe the less pedantic approach of the "additional analogies" (cut?) would help bridge that once we relax the idea insisting the analogies play true to the underlying implementation int their metaphor. I think I mentioned that in a comment I had put on them?

In any case, I'm leaning back towards reverting to your original version here.

functions-continued/variables-are-references.md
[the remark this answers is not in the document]
anselrognlie: Weird that a br is required to get the layout you want. 🤷   Is that due to special layout rules within a callout?

In the second example, it's missing an 'of' right at the start.  It should be:

In the case of `sum += 1`, by default, this line is evaluated as `sum = sum + 1`. Therefore, it works like the situation above! 1 retains its 1-ness and `sum` gets re-assigned to the new integer value.

And I'm fine with removing some of the += complexity. It seemed like an operator that could provide confusion about what really is being modified, so paring it down seems like a good idea.

I _don't_ know whether the additional function-focused examples are helpful here.  The implication of mutable and immutable values as function params are touched on later, so maybe they're not necessary?

anselrognlie (approved):
A little bit more discussion around the variables-are-references, but feel free to incorporate or ignore.

PR #62, opened 2021-02-10
Change 'breaking down the problem' examples to look like a list instead of pseudocode

jmaddox19:


approaching-a-problem/pseudocode-and-logic.md
```diff
@@ -300,13 +290,13 @@ Which of the following is an accurate description of the process in Examples 1 a
 ##### !end-question
 ##### !options
 
-* The developer started with equal parts of readable pseudocode and working Python code
+* The developer started by writing out subproblems in a numbered list, then used that list to write working Python code
```
tildeee, line 293: Is this option the answer? Because it reads as correct to me. Could you clarify what I'm missing, or could we chaneg this option to something more "definitely not the correct answer"?

tildeee (commented):
i like the changes. I like that it brings pseudocode's first step into more "make a numbered list"-- I'd appreciate if you did a "lesson flow" check to make sure that the numbered lists are mentioned (doesn't need to go into detail) at least once in the above paragraphs, and that they don't seem random.

Requesting changes because it feels like there's two valid answers to me

If that's fixed and the Learn preview works, I'm down with this

approaching-a-problem/pseudocode-and-logic.md
jmaddox19: Its missing anything about pseudocode, so it felt pretty clearly wrong to me, but please let me know what you'd suggest I change to make it more clearly wrong.

jmaddox19:
@tildeee In my mind the numbered list is mentioned because its what "breaking down the problem into subproblems" means to me. 
I will make a quick change to try to make that more explicit. Lemme know what you think

audreyandoy (approved):
Great! Thank you for updating the numbered list of subproblems, looks much better.

tildeee (approved):

PR #68, opened 2021-02-26
Practice fundamentals vocab update

tildeee:
There was feedback around some questions being confusingly worded.

While I cleaned up some language, I added a bunch of small questions for this exercise.

**It is genuinely probably faster to review this by doing learn preview.** Seriously, the questions are so small!


beccaelenzil (approved):
I completed all the exercises. They were clear to me!

PR #69, opened 2021-02-26
everywhere: various clarity fixes via feedback

tildeee:
Generally all cosmetic. Responding to collected instructor feedback.


intro-to-tests/practice-intro-to-pytest.md
```diff
@@ -864,6 +899,25 @@ class TestPython1(unittest.TestCase):
 ```
 
 ##### !end-tests
+##### !hint
```
anselrognlie, line 902: Arguably, the tests here require no more complex a solution than the previous question. If we are going to plant the TDD seed here that we really want to write the minimum code to make a test pass, but also want to encourage students to provide a real implementation here, then we should probably have both the even and odd test cases in the presented tests.

intro-to-tests/practice-intro-to-pytest.md
```diff
@@ -826,6 +859,8 @@ def is_odd(num):
     return True
 ```
 
+Note: Recall that the instructions are to write a function to make the test pass. :)
```
anselrognlie, line 862: engaging super friendly mode!
```suggestion
Does this seem too basic? Recall that the instructions were to write a function _to make the test pass_. Not to fully implement the function. 😀
```

intro-to-tests/practice-intro-to-pytest.md
```diff
@@ -963,6 +1036,22 @@ class TestPython1(unittest.TestCase):
 ```
 
 ##### !end-tests
+##### !hint
+
+Read through the test. Determine:
+
+- Which line is the "Arrange" step?
+- What are the "Act" and "Assert" steps?
```
anselrognlie, line 1044: For this problem, to me, the Arrange and Assert are what are combined rather than Act and Assert.

anselrognlie (approved):
Looks good.  I like how the hints are providing a framework for how to approach these problems.

PR #71, opened 2021-02-27
Be/one to one/many

beccaelenzil:
- added one-to-one and one-to-many vocabulary
- added a couple comprehensions questions about the Product and ShoppingCart example

### DIFF
--- a/composition/composition.md
+++ b/composition/composition.md
@@ -192,7 +192,7 @@ Note that `components` is a list of `component` objects. As usual, we use good n
 
 ## A Longer Example: Product and ShoppingCart
 
-Let's look at one more example of a one-to-many composition relationship.
+Let's look at one more example of one-to-one and one-to-many composition relationships.
 
 ```python
 class Product:
--- a/composition/composition.md
+++ b/composition/composition.md
@@ -3,13 +3,21 @@
 ## Learning Goals
 
 - Use attributes to create composition between two classes
+- Define one-to-one and on-to-many composition relationships
 - Use instance methods to access instances of objects composed inside another object
+
+## Vocabulary and Synonyms
+
+| Vocab <div style="width:100px;">| Definition |Synonym|How to Use in a Sentence |
+| ----- | ---------- |----|------------------------ |
+| One-to-one | One composite object is associated with one component object  | has-a |On an e-commerce site, each order has a *one-to-one* relationship with each shipment (destination address, tracking number, etc). <br/> In a gradebook application, each student *has a* final grade. <br/> On a social media platform, each user *has one* profile picture|
+| One-to-many | One composite object is associated with a collection of component objects. | has-many | In a gradebook application, one class *has many* students and *has many* assignments. <br/> On a social media platform, one user has a *one-to-many* relationship with posts.|
 
 ## Initializing Components as Attributes in the Constructor
 
 It's common to set the attributes of a composite class to component instances provided by other code that uses our class. Python lets us assign to attributes directly. We can let all the components in our class be passed into the constructor of the composite class, and assign the attributes in the constructor itself.
 
-Consider this example:
+Consider this example of a one-to-one composition relationship:
 
 ```python
 class ExampleComponent:
@@ -40,7 +48,7 @@ Orange has an attr named my_favorite_component: <__main__.ExampleComponent objec
 We can read the name from Orange's fav component using dot notation: apple
 ```
 
-This works similarly even with a list of instances:
+This works similarly even with a list of instances, a one-to-many composition relationship:
 
 ```python
 class ExampleComponent:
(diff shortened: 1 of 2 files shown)


composition/composition.md
```diff
@@ -151,30 +159,63 @@ Now, I'm going to leave Component's instance method
...
+class ExampleComposite:
 
-        for product in self.products:
-            total_price += product.calculate_cost()
+    def __init__(self, name, components):
+        self.name = name
+        self.components = components
+```
+
+Note that `components` is a list of `component` objects.
```
anselrognlie, line 191: consider taking the opportunity to  remind the reader about good naming
```suggestion
Note that `components` is a list of `component` objects. As usual, we use good naming, here the plural `components`, to help us remember what we expect an attribute to contain.
```

composition/composition.md
```diff
@@ -151,30 +159,63 @@ Now, I'm going to leave Component's instance method
...
+
+## A Longer Example: Product and ShoppingCart
+
+Now we will look at one more example of a one-to-many composition relationship.
```
anselrognlie, line 195: I think this sounds a little more friendly.
```suggestion
Let's look at one more example of a one-to-many composition relationship.
```

composition/composition.md
```diff
@@ -198,6 +239,96 @@ The total cost of the products in the shopping cart is $213.48
...
+
+##### !end-answer
+
+<!-- other optional sections -->
+<!-- !hint - !end-hint (markdown, hidden, students click to view) -->
+<!-- !rubric - !end-rubric (markdown, instructors can see while scoring a checkpoint) -->
+<!-- !explanation - !end-explanation (markdown, students can see after answering correctly) -->
+##### !explanation
+
+An instance of a `ShoppingCart` has many `Product`s.
```
anselrognlie, line 280: consider expanding the explanation a little
```suggestion
An instance of a `ShoppingCart` has many `Product`s. Since instances of `Product` are contained in `ShoppingCart`, `Product` is the component class, while `ShoppingCart` is the composite class.
```

composition/composition.md
```diff
@@ -151,30 +159,63 @@ Now, I'm going to leave Component's instance method
...
+```python
+1 class Product:
+2 
+3    def __init__(self, price, quantity):
+4        self.price = price
+5        self.quantity = quantity
+6
+7    def calculate_cost(self):
+8        return self.price * self.quantity
+9
```
anselrognlie, line 206: consider adding spaces to keep things in alignment. Alternatively, what about using 01, 02, 03, ...
```suggestion
1  class Product:
2 
3     def __init__(self, price, quantity):
4         self.price = price
5         self.quantity = quantity
6
7     def calculate_cost(self):
8         return self.price * self.quantity
9
```

composition/composition.md
```diff
@@ -198,6 +239,96 @@ The total cost of the products in the shopping cart is $213.48
...
+
+* type: multiple-choice
+* id: cd47699d-3533-4d5d-990c-470cd0b254c4
+* title: Composition
+<!-- * points: [1] (optional, the number of points for scoring as a checkpoint) -->
+<!-- * topics: [python, pandas] (optional the topics for analyzing points) -->
+
+##### !question
+
+In the example above, one line 21, `calculate_cost()` is
```
anselrognlie, line 302:
```suggestion
In the example above, on line 19, `calculate_cost()` is 
```

anselrognlie (approved):
Looks good to me.  Just a couple little things here and there.

composition/composition.md
```diff
@@ -151,30 +159,63 @@ Now, I'm going to leave Component's instance method
...
+10 class ShoppingCart:
+11
+12    def __init__(self, products):
+13        self.products = products
+14
+15    def calculate_total_cost(self):
+16        total_price = 0.0
+17
+18        for product in self.products:
+19            total_price += product.calculate_cost()
```
anselrognlie, line 216: Another alternative to numbering the lines could be to add an inline comment on the line you want to question about:
```suggestion
              total_price += product.calculate_cost()  # (A)
```
Then below, you could refer to this as the line marked with the (A) comment. This would also allow you to remove all the line numbering and not worry about alignment.

jmaddox19 (approved):
I suggest a quick rewording of the comment added to the code snippet but I love that suggestion to add a comment!

jmaddox19 (approved):

composition/composition.md
```diff
@@ -151,30 +159,77 @@ Now, I'm going to leave Component's instance method
...
-    def __init__(self, products):
+    def __init__(self, products, shipping_address):
         self.products = products
+        self.shipping_address = shipping_address
 
     def calculate_total_cost(self):
         total_price = 0.0
 
         for product in self.products:
+            #----Refer to this line for Check for Understanding Question 2----
```
anselrognlie, line 226: my only concern with this phrasing is that it depends on learn numbering. Personally, I prefer having the marker be something we explicitly control and reference from the question rather than depending on learn numbering the questions.

composition/composition.md
```diff
@@ -151,30 +159,77 @@ Now, I'm going to leave Component's instance method
...
             total_price += product.calculate_cost()
 
         return total_price
+
+    def.summary(self):
+        print(len(self.products),"products will be shipped to", self.shipping_address.name,".")
```
anselrognlie, line 232: suggest using interpolation (I also think technically the current code doesn't produce the presented output, since comma-separated print inserts spaces).
return a string and print it externally.
```suggestion
        return f"{len(self.products)} products will be shipped to {self.shipping_address.name}."
```

composition/composition.md
```diff
@@ -183,21 +238,121 @@ p2 = Product(2.50, 13)
...
+    state="WA",
+    zip="98112"
+    )
+
+sc = ShoppingCart(products, ravis_address)
 
 total_cost = sc.calculate_total_cost()
 
 print(f"The total cost of the products in the shopping cart is ${total_cost}")
+sc.summary()
```
anselrognlie, line 254: if the above change to return a string is used, use this as well
```suggestion
print(sc.summary())
```

anselrognlie (approved):
I like the additional content. I feel a little weird to call out question 2 directly in the example, but not enough to quibble. There are a couple minor code edits that I feel more strongly about.

PR #72, opened 2021-03-02
Lists and Memory Lesson

CheezItMan:
I'm happy to adjust this as needed.  Currently the lesson covers:

1.  How Python represents arrays/lists in memory (references to a list of references).
1.  The implications for references and functions in our code (pass by reference issues).
1.  Big O and lists
1.  A small number of exercises

~TODO~:  I'm still updating the images to meet the new [image style guide](https://docs.google.com/document/d/157HPfAaBkRScFxcBoGristaNmfL_h9iGLz8lB4liFzA/edit)

I've modified the images, but left the font in place as it would be a pain to change at this stage.


lists-and-memory/lists-as-references.md
```diff
@@ -0,0 +1,410 @@
...
+
+## Goal
+
+- Explain the impact of lists as references in a function call
+- Identify functions with side effects
+- Explain reasons to avoid side effects in our
+
+## Introduction
+
+Lists are our first data structure.  As we saw in the last lesson a list is a more sophisticated data structure storing references to each element indirectly.  In this lesson we will look at some of the implications we need to keep in mind as developers.
```
beccaelenzil, line 11:
```suggestion
Lists are our first data structure.  As we saw in the last lesson, a list is a more sophisticated data structure storing references to each element indirectly.  In this lesson we will look at some of the implications we need to keep in mind as developers.
```

lists-and-memory/lists-as-references.md
```diff
@@ -0,0 +1,410 @@
...
+
+## Vocabulary and Synonyms
+
+| Vocab | Definition | Synonyms | How to Use in a Sentence
+| --- | --- | --- | ---
+|  Side Effect | Any lasting effect that occurs in a function, which is not through its return value. | - | A side-effect of our `find_product` function is that our list of products is put into alphabetical order. |
+|  Mutable data type | Any type of variable which can be modified without changing the reference. | changable object | Lists and dictionaries can be modified inside function calls and so are mutable. |
```
beccaelenzil, line 18:
```suggestion
|  Mutable data type | Any type of variable which can be modified without changing the reference. | changeable object | Lists and dictionaries can be modified inside function calls and so are mutable. |
```

lists-and-memory/lists-as-references.md
```diff
@@ -0,0 +1,410 @@
...
+
+print(f"Before calculating the the total of the order, products_total is \
+{products_total}")
+bill_total = calculate_subtotal(products_total, 0.08)
+print(f"The total bill is {bill_total}")
+print(f"After calculating the the total of the order, products_total is \
+{products_total}")
+```
+
+While the function `calculate_total` is running both the parameter `subtotal` and the argument `products_total` both refer to the same value in memory.
```
beccaelenzil, line 41:
```suggestion
While the function `calculate_total` is running, the parameter `subtotal` and the argument `products_total` both refer to the same value in memory.  
```

lists-and-memory/lists-as-references.md
```diff
@@ -0,0 +1,410 @@
...
+print(f"The total bill is {bill_total}")
+print(f"After calculating the the total of the order, products_total is \
+{products_total}")
+```
+
+While the function `calculate_total` is running both the parameter `subtotal` and the argument `products_total` both refer to the same value in memory.  
+
+![Integer references from above](../assets/lists-and-memory/lists-and-memory_lists-as-references_int-references.png)
+
+*Fig. A variable in Python contains a reference to the location where the data is stored.
```
beccaelenzil, line 45:
```suggestion
*Fig. A variable in Python contains a reference to the location where the data is stored.*
```

lists-and-memory/lists-as-references.md
```diff
@@ -0,0 +1,410 @@
...
+
+
+**Question:** When the line: `subtotal += subtotal * tax_rate` executes, subtotal changes in value.  Why doesn't `products_total` change?
```
beccaelenzil, line 48:
```suggestion
**Question:** When the line: `subtotal += subtotal * tax_rate` executes, `subtotal` changes in value.  Why doesn't `products_total` change?
```

lists-and-memory/lists-as-references.md
```diff
@@ -0,0 +1,410 @@
...
+        shortened_names.append(name[0:5])
+    
+    return shortened_names
+```
+
+## Summary
+
+Because lists internally contain references to the elements stored in them, it is possible for a parameter to modify the elements of a list passed as a parameter.  When this happens it is called a _side effect_.  Side effects can be useful, but they can also be problematic because they make our programs harder to test, and provide an avenue for subtle and hard to diagnose bugs enter our applications.  In general we try to avoid introducing side effects.
+
+## Learning Comprehension
```
beccaelenzil, line 145:
```suggestion
## Check for Understanding
```

lists-and-memory/lists-and-big-o.md
```diff
@@ -0,0 +1,172 @@
...
+      "name": name,
+      "class": class_name,
+    })
+
+    return students
+```
+
+The above function takes a list and a couple of arguments, but the function does not take any longer to execute as the list grows larger.
+
+You might think that traversing a list, means that a function's time complexity is usually going to be O(n), but this is not always the case.
```
beccaelenzil, line 41:
```suggestion
We might think that traversing a list, means that a function's time complexity is usually going to be O(n), but this is not always the case.  
```

lists-and-memory/lists-and-big-o.md
```diff
@@ -0,0 +1,172 @@
...
+    for student in students:
+        if student["name"] == student_name:
+            return student
+
+    return None
+```
+
+The function takes an array as input, but it does not create a new list, and it does not increase the size of the given list.  So this function's space complexity is O(1).
+
+Similarly the function below only ever increases the size of the list by one element.  It always adds one dictionary with two key-value pairs, no matter the size of the input.
```
beccaelenzil, line 97:
```suggestion
Similarly the function below only ever increases the size of the list by one element.  It always adds one dictionary with two key-value pairs, no matter the size of the input. So this function's space complexity is also O(1).
```

lists-and-memory/lists-and-big-o.md
```diff
@@ -0,0 +1,172 @@
...
+def add_student(students, name, class_name):
+    students.append({
+      "name": name,
+      "class": class_name,
+    })
+
+    return students
+```
+
+On the other hand this function **does** allocate more memory as the size of the input increases.
```
beccaelenzil, line 109:
```suggestion
On the other hand, this function **does** allocate more memory as the size of the input increases.
```

lists-and-memory/lists-and-big-o.md
```diff
@@ -0,0 +1,172 @@
...
+        How can you identify the time or space complexity of a Python library function?
+    </summary>
+
+To identify the time/space complexity of a Python library function you could:
+
+* Read the [Python Documentation entry on it](https://docs.python.org/3/tutorial/datastructures.html)
+* Search for the answer in a web search engine (google it)
+* Ask someone either in person or on a site like Stack Overflow
+
+Regardless you may need to make some educated guesses on the function's performance based on the description you read.  Consider, "How would I have to do this?"
```
beccaelenzil, line 167:
```suggestion
        How can we identify the time or space complexity of a Python library function?
    </summary>

To identify the time/space complexity of a Python library function we can:

* Read the [Python Documentation entry on it](https://docs.python.org/3/tutorial/datastructures.html)
* Search for the answer in a web search engine (google it)
* Ask someone either in person or on a site like Stack Overflow

Regardless, we may need to make some educated guesses on the function's performance based on the description we read.  Consider, "How would I have to do this?"
```

beccaelenzil (approved):
Lots of small edits, and changing you and your to we and our. Love the lessons!

PR #77, opened 2021-03-04
embedded videos to dictionaries and lists

spitsfire:
Checked videos in preview. Looks good.


anselrognlie (approved):
🦥  thinks it looks great!

PR #81, opened 2021-03-04
Be/conditional problem reword

beccaelenzil:
I reworked problem 7 to remove compound conditionals and be about the order of the conditional tests.

### DIFF
--- a/fundamentals-recap/problem-set-conditionals-recap.md
+++ b/fundamentals-recap/problem-set-conditionals-recap.md
@@ -458,8 +458,6 @@ if customer_age < 60:
     ticket_price = 15.00
 if customer_age >= 60:
     ticket_price = 11.00
-else:
-    print("please enter a valid age.")
 ```
 * 
 ``` Python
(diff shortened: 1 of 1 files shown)


fundamentals-recap/problem-set-conditionals-recap.md
```diff
@@ -492,13 +500,15 @@ else:
 
 * 
 ``` Python
-if customer_age <= 10:
+if customer_age < 0:
```
anselrognlie, line 503: Whatever change is applied to the correct answer above should be mirrored here (either flipping => to >= or changing to an else).

anselrognlie (approved):
The >= operator is backwards in a few places (hard to break the rocket arrow habit!) 😀 

For the correct answer, I think the final else is unnecessary and the final elif could be replaced with else on its own. But either replacing the => with >= or the else replacement _should_ be ok.  I'll leave it to you.

(full disclosure, even when writing this comment I originally wrote <= instead of >=. It looks so wrong! 😂 )

PR #82, opened 2021-03-05
ar/problem set while loops

anselrognlie:
Adds problem set for while loops
the problem breakdown is
- how many times will this loop run (multiple choice x 5) to test understanding of the various parts of a while loop and associated keywords
- while/for loop equivalency (multiple choice x 3) to reinforce that they can do some of the same things, but maybe one is nicer than the other for certain situations
- reorder loop lines (x1) to show understanding of the difference between continue and break
- short coding challenges (x2) to use a while loop for looping over data out of order and looping while waiting for multiple conditions and user input
- pick appropriate scenarios for while vs. for loops (checkbox x 1) to synthesize their work from lessons and the problem set into thinking about when to use one type of a loop over the other

`sidewinder` coding challenge includes a diagram to help with the explanation


while-loops/problem-set-while-loops.md
```diff
@@ -0,0 +1,849 @@
...
+### !challenge
+
+* type: code-snippet
+* language: python3.6
+* id: 85dffbd3-2030-44b4-8b9f-cc86f0c7e9d1
+* title: While Loops
+
+##### !question
+
+Let's implement a function called `sidewinder`. It will take a parameter `number_list` which contains a list of the _next_ index in the list to visit. Starting from position `0`, we will look in the current position, get the value stored there, and use it as the _next_ position to visit. We will continue doing this until we find a value of `0`, which means we have found the end of the index chain. Then we'll return the position where we found the `0`.
```
tildeee, line 539: I'm making a suggestion for how to re-word this function. Summary of recommendations:

- Lead the description with what the function is responsible for doing ultimately, even before details of how to get there
- Makes up verbiage that says this is a fictional algorithm, to help reassure that this is totally a made up algorithm. I didn't really do my research to find out how true that is, but I would want to assure students that this is something made up for an exercise, not an algorithm that they should memorize.
- `number_list` _is_ a list of those indices (not contains)
- Words "starting from position `0`" as a rule necessary to solving the problem, rather implies it
- Formats it as a bulleted list to help students turn it into code
- I started to use the word "index" so I added a parenthetical (index) to the final use of "position"

```suggestion
Let's implement a function called `sidewinder`. This function is responsible for finding the _index_ in a list that has the value `0` when following the imaginary "sidewinder algorithm."

In order to find the correct index, the function:
  - Takes a parameter `number_list` which is a list of "next" indices to visit
  - The function always starts at index `0`
  - The function will look in the current position, get the value stored there. Then, it will use that value as the _next_ position to visit
  
The function continues doing this until we find a value of `0`, which means we have found the end of the index chain. Then we'll return the _position_ (index) where we found the `0`.
```

while-loops/problem-set-while-loops.md
```diff
@@ -0,0 +1,849 @@
...
+<div style="max-width: 700px; margin: 0 auto -28px auto; text-align: center;">
+<img alt="How to interpret the array input to sidewinder" src="../assets/while-loops_problem-set-while-loops_sidewinder.png">
+</div>
+*Fig. Given the list `[4, 9, 0, 0, 1, 8, 0, 0, 11, 5, 0, 10]` for `number_list`, this diagram shows how to move through the list values to find the 'end.' Starting from index `0`, we find the value `4`. We go to index `4` where we find the value `1`. We go to index `1` where we find the value `9`, and so on. Eventually we reach index `10` where we find the value `0`, so this is the end and the result is `10`.*
+
+Provide an implementation for `sidewinder` so that it
+- follows the chain of indices to the end, and
+- returns the position of the end of the chain
+
+You may assume that `number_list` will never be an empty list, and that all of the values in the list are valid indices in the list.
```
tildeee, line 554: Nitpick that I'm not married to: I tend to write "Assume," not "You may assume," so for consistency...

```suggestion
Assume that `number_list` will never be an empty list, and that all of the values in the list are valid indices in the list.
```

while-loops/problem-set-while-loops.md
```diff
@@ -0,0 +1,849 @@
...
+
+def input_int(*args):
+    return next(num_src)
+
+# inject into main
+main.input_int = input_int
+
+class TestPython1(unittest.TestCase):
+
+    def test_one(self):
```
tildeee, line 709: My ideal is that the test names are a bit more descriptive. When debugging, students can see these test names as the tests that fail, so it's valuable to them.

```suggestion
    def test_silly_sum_stops_at_val_zero(self):
```

while-loops/problem-set-while-loops.md
```diff
@@ -0,0 +1,849 @@
...
+        global num_src
+        num_src = input_ints([1, 1, 2, 3, 5, 8, 0, 9])
+        self.assertEqual(20, silly_sum(), 'the provided input was 1, 1, 2, 3, 5, 8, 0, 9')
+
+    def test_two(self):
```
tildeee, line 714: Mentioned above

```suggestion
    def test_silly_sum_stops_after_sum_is_greater_than_1000(self):
```

while-loops/problem-set-while-loops.md
```diff
@@ -0,0 +1,849 @@
...
+        global num_src
+        num_src = input_ints([400, 400, 400, 400])
+        self.assertEqual(1200, silly_sum(), 'the provided input was 400, 400, 400, 400')
+
+    def test_three(self):
```
tildeee, line 719:
```suggestion
    def test_silly_sum_stops_when_sum_is_1001(self):
```

while-loops/problem-set-while-loops.md
```diff
@@ -0,0 +1,849 @@
...
+from main import sidewinder
+
+class TestPython1(unittest.TestCase):
+  def test_one(self):
+    self.assertEqual(0, sidewinder([0]))
+
+  def test_two(self):
+    self.assertEqual(1, sidewinder([2, 0, 1]))
+
+  def test_three(self):
```
tildeee, line 580: Students can see these test names when debugging, so it's helpful to rename them

```suggestion
  def test_sidewinder_when_list_of_one_element_zero(self):
    self.assertEqual(0, sidewinder([0]))

  def test_sidewinder_stops_at_value_zero(self):
    self.assertEqual(1, sidewinder([2, 0, 1]))

  def test_stops_at_value_zero_in_long_list(self):
```

while-loops/problem-set-while-loops.md
```diff
@@ -0,0 +1,849 @@
...
+<img alt="How to interpret the array input to sidewinder" src="../assets/while-loops_problem-set-while-loops_sidewinder.png">
+</div>
+*Fig. Given the list `[4, 9, 0, 0, 1, 8, 0, 0, 11, 5, 0, 10]` for `number_list`, this diagram shows how to move through the list values to find the 'end.' Starting from index `0`, we find the value `4`. We go to index `4` where we find the value `1`. We go to index `1` where we find the value `9`, and so on. Eventually we reach index `10` where we find the value `0`, so this is the end and the result is `10`.*
+
+Provide an implementation for `sidewinder` so that it
+- follows the chain of indices to the end, and
+- returns the position of the end of the chain
+
+You may assume that `number_list` will never be an empty list, and that all of the values in the list are valid indices in the list.
+
```
tildeee, line 555: I've personally enjoyed and treasured providing the tests in `pytest` form whenever there's a code challenge. Is it possible for you to do that too?

while-loops/problem-set-while-loops.md
```diff
@@ -0,0 +1,849 @@
...
+##### !question
+
+Let's sum some numbers! We will write a function called `silly_sum` that reads a series of numbers from user input and returns the sum. To keep things interesting, let's also decide that entering `0` means the user is done entering input, and that we'll also stop asking for further input if the sum reaches or exceeds `1000`.
+
+Provide an implementation for `silly_sum` so that it
+- reads numbers from the user (use `input_int`) summing as we go until either
+    - the user enters `0`, or
+    - the sum reaches or exceeds `1000`
+
```
tildeee, line 673: I've personally enjoyed and treasured providing the tests in `pytest` form whenever there's a code challenge. Is it possible for you to do that too?

while-loops/problem-set-while-loops.md
```diff
@@ -0,0 +1,849 @@
...
+    # stores the most recent user input
+    # default to anything except 0 (think about why this is)
+    num = 1
+
+    # check our 2 conditions:
+    #   user input must not be 0
+    #   don't ask for more input once the sum reaches at least 1000
+    while num != 0 and my_sum < 1000:
+        # prompt for input
+        num = input_int('Enter an integer (0 to stop): ')
```
tildeee, line 769: From the information provided, I didn't think that I should add a string to `input_int`. If this is something we expect the students to have been able to organically write, I would change something about the intro to `input_int` or not include this arg?

while-loops/problem-set-while-loops.md
```diff
@@ -0,0 +1,849 @@
...
+
+Read through this version and compare to yours. What's the same? What's different? Are there even more ways to approach this problem?
+
+<br />
+
+For some extra challenge, on your own, try modifying `silly_sum` so that you can pass in the stop value and maximum sum as parameters. This would allow us to say that 1 is the stop value instead of 0. How does this affect the code? Does anything else need to change? We could also say that the maximum sum is -1000! What changes does that cause?  
+
+<br />
+
+Simple changes can sometimes have complicated effects. Play around a bit and see what you can build!
```
tildeee, line 786: As mentioned with the other question-- You're adding an extension to this question to the explanation. This is fine, but I can imagine it being helpful to separate this and call it out as an extension/optional. I won't fight for this.

while-loops/problem-set-while-loops.md
```diff
@@ -0,0 +1,849 @@
...
+An example of a working implementation:
+
+```python
+def silly_sum():
+    # accumulate the sum
+    my_sum = 0
+
+    # stores the most recent user input
+    # default to anything except 0 (think about why this is)
+    num = 1
```
tildeee, line 762: I won't fight your implementation to change in Learn since it's ultimately a good conversation starter, but why not

```python
num = input_int()
```

?

while-loops/problem-set-while-loops.md
```diff
@@ -0,0 +1,849 @@
...
+from main import silly_sum
+
+num_src = None
+
+def input_ints(nums):
+    for num in nums:
+        yield num
+
+def input_int(*args):
+    return next(num_src)
```
tildeee, line 702: Idk if there's something actionable to do here, but I'd love to call out that while I was debugging my solution to this, my stacktrace looked like this, which was specifically _not_ helpful to me lol.

```
.EE
======================================================================
ERROR: test_three (test_methods.TestPython1)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/src/app/test_methods.py", line 37, in test_three
    self.assertEqual(1000, silly_sum(), 'the provided input was 999, 1, 1')
  File "/usr/src/app/main.py", line 9, in silly_sum
    user_input = input_int()
  File "/usr/src/app/test_methods.py", line 17, in input_int
    return next(num_src)
StopIteration

======================================================================
ERROR: test_two (test_methods.TestPython1)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/src/app/test_methods.py", line 32, in test_two
    self.assertEqual(1200, silly_sum(), 'the provided input was 400, 400, 400, 400')
  File "/usr/src/app/main.py", line 9, in silly_sum
    user_input = input_int()
  File "/usr/src/app/test_methods.py", line 17, in input_int
    return next(num_src)
StopIteration

----------------------------------------------------------------------
Ran 3 tests in 0.001s

FAILED (errors=2)
```

while-loops/problem-set-while-loops.md
```diff
@@ -0,0 +1,849 @@
...
+<!-- ======================= END CHALLENGE ======================= -->
+<!-- prettier-ignore-end -->
+
+### A Little Helper for the Next Question
+
+The following coding question uses a helper function called `input_int`. It works just like the [`input`](https://docs.python.org/3/library/functions.html#input) function that we've seen before, except that instead of returning a string, we will assume that it returns an integer.
+
+In the question block below, we don't need to define `input_int`. We can assume that it is available, the same way `input` is available.
+
+To experiment with the coding question in VS Code or repl.it, we will need an implementation for `input_int`. In Python, safely converting from a string to an integer requires handling exceptions. We haven't discussed that material yet, so here is a working implementation.
```
tildeee, line 633: I'm suggesting some wording changes to help angle this moreso from "Here's stuff you need to know for the next question to," "The next question is special because we're providing this. You should use it for the next question," in an attempt to strongly state "what you need is to use it, not to know it"

Summary of suggestions:

- Changes header (and content) tone to emphasize that it's provided and it's for the next question
- Changes tone to "it's already there, this is an FYI"
- Surfaces the responsibility of this function/why the student should care about this function first
- Give context to why it matters that we shouldn't define it and why we can assume it's available-- it's because students should use it
- removes the assumption that 1. this lesson always comes before handling exceptions and 2. that students haven't independently learned about that anyway

```suggestion
### Provided Helper Function for the Next Question

In the following coding question, there is a helper function called `input_int`. This function is responsible for getting user input, and it will always return an integer. (It works similarly to the [`input`](https://docs.python.org/3/library/functions.html#input) function that we've seen before, except the function does not return a string.)

In the question block below, `input_int` is already defined and available to use in our code. Do not define `input_int`. Use it because it is available, the same way `input` is available.

If you would like to experiment with the coding question in VS Code or repl.it, you will need an implementation for `input_int`. Here is a working implementation for you to copy/paste and use in VS Code or repl.it. It may contain concepts not covered yet, such as handling exceptions.
```

while-loops/problem-set-while-loops.md
```diff
@@ -0,0 +1,849 @@
...
+### !challenge
+
+* type: code-snippet
+* language: python3.6
+* id: 0fdaf2f7-0573-4e0a-8fd2-46f15076de24
+* title: While Loops
+
+##### !question
+
+Let's sum some numbers! We will write a function called `silly_sum` that reads a series of numbers from user input and returns the sum. To keep things interesting, let's also decide that entering `0` means the user is done entering input, and that we'll also stop asking for further input if the sum reaches or exceeds `1000`.
```
tildeee, line 667: I'm personally not a fan of the phrase "to keep things interesting" because... they're just added requirements :P

I won't fight for taking it out, but I think that the rest of the question speaks for itself

```suggestion
Let's sum some numbers! We will write a function called `silly_sum` that reads a series of numbers from user input and returns the sum while following the rules below.
```

tildeee (approved):
The questions work and look good enough as-is, and it wouldn't be bad if it were merged. However, I have a bunch of thoughts and questions about style, etc etc etc

while-loops/problem-set-while-loops.md
```diff
@@ -0,0 +1,849 @@
...
+
+##### !answer
+
+1. while True:
+1.     user_input = input('enter input: ')
+1.     if user_input == 'done':
+1.         break
+1.     elif user_input == 'skip':
+1.         continue
+1.     print(f'you entered {user_input}')
```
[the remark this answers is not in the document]
anselrognlie: No, it wouldn't. I _did_ do that intentionally, to make them pay attention. But if you feel that's more confusing than engaging, I'm fine with changing it.

while-loops/problem-set-while-loops.md
```diff
@@ -0,0 +1,849 @@
...
+
+### !challenge
+
+* type: checkbox
+* id: f44b4f27-82e7-4545-a4d0-766e307971b1
+* title: While Loops
+
+##### !question
+
+Combining what we've learned about `while` loops and `for` loops both from the previous lessons and the previous questions, which of the following situations are handled more appropriately with a `while` loop?
```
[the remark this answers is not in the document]
anselrognlie: Aw.  I thought I did a good since I was patterning this after some verbiage that has come up in previous reviews I remember doing. 😭

while-loops/problem-set-while-loops.md
anselrognlie: I don't like repeating the core part of the logic (fetching user input) if I can avoid it (especially as part of what amounts to just a variable declaration). Honestly, it kills me that Python doesn't have a do/while.

Personally, I wouldn't even use num in this implementation.  I'd just start up a while True and break as necessary!

I'm inclined to leave this as is.

while-loops/problem-set-while-loops.md
anselrognlie: The test is simulating user input, and the code didn't stop reading it before reaching the end, which results in an uncaught StopIteration error.  The snowman project ran into this as well...

I _could_ add an except and translate it into a custom error that maybe reads more intuitively as a YouReadTooManyThingsTryAddingTerminateConditionsError (not this seriously, but something more descriptive than the built int StopIteration).

Thoughts?

while-loops/problem-set-while-loops.md
anselrognlie: I just know that these are dumb, unrealistic requirements, and wanted to acknowledge that to the students.  Try to stave off the inevitable "why would anyone _do_ this?" response.

But happy to cut it if you feel it's unnecessary.

while-loops/problem-set-while-loops.md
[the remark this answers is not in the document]
tildeee: I'm into the "confusing-ness" of it knowing that it was intentional, keep it ;)

while-loops/problem-set-while-loops.md
tildeee: replit.com would future-proof us, yeah!!!!

I'm manually editing my suggestion above to that, too, in case that's helpful.

while-loops/problem-set-while-loops.md
tildeee: I personally feel like I don't have enough data to think of the "perfect" solution...

As a "sick and twisted" educator, instead of fool-proofing this, I think I propose:

- Doing a watered-down "catch all" hint: Make a hint that says "If you see a `StopIteration` error, it is because ... . This is likely happened because the function's logic did not stop before reaching the end of the list. To resolve this error, look for logical errors again. Try pairing and rubber duck debugging if you need to!"
- Possibly adding the smallest `.instructor.md` resource here and make a 2 second explanation about this (aka acknowledge that `StopIteration` error could exist)

Hopefully by C16, we'll have collective wisdom for how to _actually_ deal with this? Maybe?


Happy for other suggestions, too :D The suggestion I just gave is essentially "I don't know how to suggest to rewrite the question" looool


PS: For more data, I got this error because I checked for the number to be greater than `1000`, instead of the sum... purely a logical error, not syntax one!

while-loops/problem-set-while-loops.md
tildeee: Makes sense to me! :)

while-loops/problem-set-while-loops.md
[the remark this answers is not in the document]
tildeee: If we weren't walking contradictions, would we really be human? 😆 

I won't fight for my suggestion beyond what came to mind when I read it :P The meaning of the question still comes across perfectly well.

while-loops/problem-set-while-loops.md
```diff
@@ -0,0 +1,849 @@
...
+
+Read through this version and compare to yours. What's the same? What's different? Are there even more ways to approach this problem?
+
+<br />
+
+For extra challenge, consider how we might handle bad inputs, like a missing or empty list. What if the list _does_ contain values that aren't valid indices? Or what if there's a loop in the indices like in `[1, 2, 1]`. Can you write a version of `sidewinder` which instead of returning the end of the chain, returns `True` if it visits _every_ position in the list, and `False` otherwise?
+
+<br />
+
+Be creative!
```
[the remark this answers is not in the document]
anselrognlie: Yeah. Definitely wanted to keep it hidden, and if people elect not to read the explanation and aren't burdened with seeing "yet more stuff to do" I think that's how I'd like this to be used. I wouldn't expect anyone to sit down and do this (I wouldn't require any of this additional thought to be part of a later activity), but I wanted to provide additional structure for how to take things further for anyone who might be interested.

while-loops/problem-set-while-loops.md
anselrognlie: Same as above.If people completely ignore this, fine. I just want there to be some framework for further consideration for those who might want to.

while-loops/problem-set-while-loops.md
```diff
@@ -0,0 +1,849 @@
...
+    # iterate as long as the next position value is not 0
+    while number_list[next_position] != 0:
+        # update the next position value
+        next_position = number_list[next_position]
+
+    # return the position of the end of the chain
+    return next_position
+```
+
+Read through this version and compare to yours. What's the same? What's different? Are there even more ways to approach this problem?
```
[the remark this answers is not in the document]
anselrognlie: Cool. I wanted to try to maintain the conversation.  I feel like asking the comparison questions gently reinforces the idea that there's more than one possible approach and there's value in considering the differences.

while-loops/problem-set-while-loops.md
anselrognlie: I started doing that for sidewinder, but then it really didn't make sense to do so for silly_sum since most of that is related to mocking the behavior of Input_int and just added a lot of noise. Since it didn't really work well for both, it seemed best to leave it off for both. That's why I added the message to the test error to give a sense of what was going on, while in sidewinder, we can see explicitly what list it's called with.

So I'd rather not add the explicit tests here.

while-loops/problem-set-while-loops.md
anselrognlie: From my response above, I'm not really inclined to add the tests, it _really_ doesn't help for silly_sum, and it felt weird to do it for one but not the other.

while-loops/problem-set-while-loops.md
anselrognlie: Do we not show using `input` that way anywhere?!??!!?

while-loops/problem-set-while-loops.md
```diff
@@ -0,0 +1,849 @@
...
+* type: code-snippet
+* language: python3.6
+* id: 85dffbd3-2030-44b4-8b9f-cc86f0c7e9d1
+* title: While Loops
+
+##### !question
+
+Let's implement a function called `sidewinder`. It will take a parameter `number_list` which contains a list of the _next_ index in the list to visit. Starting from position `0`, we will look in the current position, get the value stored there, and use it as the _next_ position to visit. We will continue doing this until we find a value of `0`, which means we have found the end of the index chain. Then we'll return the position where we found the `0`.
+
+For example, if we received `[2, 0, 1]` as our `number_list`, starting from position `0` we find a `2`. We look in position `2` and find a `1`. So we look in position `1` and find `0`. `0` marks the end of the chain, so we return `1`, the position we just looked in.
```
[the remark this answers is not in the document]
anselrognlie: I think I can add another diagram. It's funny how complicated it is to explain this problem when the implementation is so tiny.

while-loops/problem-set-while-loops.md
anselrognlie: tanx

while-loops/problem-set-while-loops.md
anselrognlie: Love it! Just changed "when following" to "by following"

```suggestion
Let's implement a function called `sidewinder`. This function is responsible for finding the _index_ in a list that has the value `0` by following the imaginary "sidewinder algorithm."

In order to find the correct index, the function:
  - Takes a parameter `number_list` which is a list of "next" indices to visit
  - The function always starts at index `0`
  - The function will look in the current position, get the value stored there. Then, it will use that value as the _next_ position to visit
  
The function continues doing this until we find a value of `0`, which means we have found the end of the index chain. Then we'll return the _position_ (index) where we found the `0`.
```

while-loops/problem-set-while-loops.md
anselrognlie: The hint tag is currently used to show the possible inputs that the tests are looking for, so I think I'd prefer to try raising a more descriptive error first.  If it doesn't work well, I'll fall back on sticking more info in the hint.

An instructor file just in case sounds good too. That'll be my first one!

while-loops/problem-set-while-loops.md
anselrognlie: I'm seeing cases where we show `input` being used with a prompt in our coursework, so I think this should be ok.  The test _doesn't_ check for any output at all (either for a prompt or for a result), so I feel okay with leaving the prompt as a nicety in the solution.

while-loops/problem-set-while-loops.md
anselrognlie: Now if you go past the end of input, you will get
`test_methods.ReadTooManyInputsError: Your loop has read beyond the end of the user input!`

While I do like the thought of having students dig into investigating the StopIteration, in all fairness, libraries _do_ try to report _useful_ exceptions rather than making us dig into their implementations.  If we were covering generators, then StopIteration would be great, but since it's really an implementation detail here, and students will have already slogged through a pretty long set of questions, I feel like taking it down a notch and just saying what really went wrong is reasonable.

PR #83, opened 2021-03-05
reversed precedence, and made examples without ()

beccaelenzil:
The precedence list was in the reverse order, and parentheses were used in all the examples. I reversed the precedence list, and wrote examples without parentheses.


CheezItMan (approved):
Looks good, merging!

PR #84, opened 2021-03-06
ar/replaces diagram image in nested loops.

anselrognlie:
Also adds a caption to the diagram.


anselrognlie:
Got offline sign off from @jbieniosek

PR #87, opened 2021-03-09
Revises explanation of chained relational operators

anselrognlie:
Python handling of expressions with multiple relational operations is
different from many other languages. The previous example would be
correct for Ruby, JS, C/C++, Java... bunches of others.

This change clearly marks the old content as incorrect (in an effort
to be fully transparent with the students) and directs
students to the new content added later, in an ordering determined
by needing to have an understanding of operator precendence
before presenting the easiest way to understand (roughly) what
Python is doing.

An additional header is also added to separate the description of how
parentheses affect expression evaluation, from a series of example
expressions, none of which use parentheses.

An extra space and a misspelling is also cleaned up in those examples.

### DIFF
--- a/fundamentals-recap/conditionals-recap.md
+++ b/fundamentals-recap/conditionals-recap.md
@@ -145,7 +145,7 @@ Generally, conditional expressions are evaluated from left-to-right, and each ex
 
 ### !callout-danger
 
-## Errata Notice
+## Correction Notice
 The following crossed-out content is incorrect in Python, but _does_ apply to many other languages. For a fuller explanation of interpreting conditional expressions with multiple relational operators in Python, refer to the _Chaining Relational Operators of Equal Priority_ section later in this lesson.
 
 ### !end-callout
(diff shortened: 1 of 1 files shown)


CheezItMan (approved):
That explains quite a bit @anselrognlie .  I think we can roll this into Learn and it would make a great topic for a roundtable discussion.

fundamentals-recap/conditionals-recap.md
```diff
@@ -143,6 +143,15 @@ Conditional expressions can grow in size and complexity. How do we determine tha
 
 Generally, conditional expressions are evaluated from left-to-right, and each expression with one operator resolves first before moving to the next operator.
 
+### !callout-danger
+
+## Errata Notice
```
kaidamasaki, line 148: I feel like "errata" is kind of an obscure word that not all students might be familiar with.

Maybe:
```suggestion
## Correction Notice
```

beccaelenzil (approved):

kaidamasaki (approved):
Looks good to me.

tildeee (approved):

PR #88, opened 2021-03-11
Aa/visual work

audreyandoy:
Fixing column widths to:

1. Representing Data in Lists and Dictionaries 
2. Iterating over Lists
3. Iterating over Dictionaries
4. Iterating and Special Cases

Adding images to:
1.  Representing Data in Lists Dictionaries
2.  Iterating over Dictionaries


representing-data/representing-data-in-lists-and-dictionaries.md
```diff
@@ -108,11 +111,11 @@ To practice representing data as lists, we will...
...
+| Data must be organized in key-value pairs | All data in dictionaries _must_ be in key-value pairs; there is no way to add a value without specifying a key                                                                                      |  ``` {"key_name": "value", "another_key": 100} ```                            |
+| Data is unordered                         | Dictionaries may sometimes appear to have an order, but there is no consistent ordering of key-value pairs. This is helpful to express that there isn't necessarily priority or importance to order | ``` {"key_123": 10,    "key_200": 0.34,   "key_2": 1}``` Notice how despite there being a numbering convention in these key-value pairs that they are not in a specific order.  |
+| Accessing values by using keys            | All values can be accessed by a key; this might have more meaning than accessing values by index, especially when keys are named intentionally                                                      | We can access values by using square brackets, ```dict_name["key_name"]```    |
```
tildeee, line 118: I previewed this and it's totally good! No request for change!

Also, I'm curious about this being formatted as a code block (which uses the triple backtics) as opposed to an inline code snippet (which uses single backtics)?

tildeee (approved):
I love it!!! Totally amazing

I know this sucks and it's a quick turn around, but since these changes are in upcoming lessons, we should merge this ASAP

PR #97, opened 2021-03-17
Don't highlight stack traces as bash

kaidamasaki:
We might want to wait until C16 to merge this but I figured having the PR could be handy.


kaidamasaki:
I made a "Cohort 16" label in case we want to stick that on things we want to hold off on merging.

CheezItMan (approved):
I don't see any downside to merging this in.  We don't really lose anything and I don't see it confusing students.

PR #107, opened 2021-03-22
Adding function call to problem #2

jbieniosek:


anselrognlie (approved):
Yay!

Out of scope for this PR, but I want to say that I wish fewer of our examples side-effected the input.

PR #109, opened 2021-03-22
Fix misspelling of ZeroDivisionError

kaidamasaki:


tildeee (approved):
wow, one of those typos that my eyes can't detect without squinting. :P cool

kaidamasaki:
As a group we caught it in office hours because a student copy/pasted the exception name.

PR #121, opened 2021-04-02
Quick fix to inaccurate Answer

jmaddox19:
I fixed the answer to the original question to make it accurate, then realized it'd probably be nice to have a 2nd question where we actually do show how Bear could inherit Mammal's attributes.


jmaddox19:
Closing PR for us to merge in at the end of C15 to not disrupt the answers to the problem set

PR #126, opened 2021-04-27
refactor classes to using dicts

tgoslee:


intermediate-python-debugging/debugging-with-vs-code.md
```diff
@@ -95,16 +95,16 @@ Notice that the program will pause at the first breakpoint it encounters.
 
 You can create a watch by clicking on the `+` icon and enter any expression.
 
-![Entering a Watch](../assets/vs-code-debugger/watch-price.png)
+![Entering a Watch](../assets/vs-code-debugger/watch-guess-before.png)
```
anselrognlie, line 98: Nice catch to add the trailing spaces!

intermediate-python-debugging/debugging-with-vs-code.md
```diff
@@ -95,16 +95,16 @@ Notice that the program will pause at the first breakpoint it encounters.
 
 You can create a watch by clicking on the `+` icon and enter any expression.
 
-![Entering a Watch](../assets/vs-code-debugger/watch-price.png)
+![Entering a Watch](../assets/vs-code-debugger/watch-guess-before.png)  
 *Fig. Creating a watch*
 
-![Watches](../assets/vs-code-debugger/watch.png)
+![Watches](../assets/vs-code-debugger/watch-guess-after.png)
```
anselrognlie, line 101: I also like that you updated these with examples that make more sense given the sample code.

intermediate-python-debugging/problem-set-using-the-debugger.md
```diff
@@ -2,7 +2,7 @@
 
 ## Directions
 
-In this problem set we will use the debugger to identify two errors.  To run the activity clone the code at [AdaGold/python-debugging](https://github.com/AdaGold/python-debugging).  Then follow the directions in the README to set up the repo.
+In this problem set we will use the debugger to identify two errors.  To run the activity clone the code at [AdaGold/python-debugging](https://github.com/AdaGold/python-debugging/tree/cm/dictionary-version).  Then follow the directions in the README to set up the repo.
```
anselrognlie, line 5: Rather than pointing at the branch, we should merge the branch into main in AdaGold. Students haven't seen how to fetch a branch from a repo.

Could you turn the branch into a PR?

From looking at the branch, so far I would say:
- remove .pylint from the repo
- remove __init__.pyc from the repo (I've never seen this, usually byte code gets put in __pycache__ dirs)
- add enough blanks (I think just one) to the top of order.py so that the line numbers line up with the  error text in this lesson (which will also require updating running-debugger.png)--there's another screenshot that shows the breakpoint still on line 16.

anselrognlie: Could you also replace `__init__` on line 79 with `create_product`
And as long as you're in this file, could you replace `Debug Test` on line 106 with <code>Debug Test</code>
It doesn't look like learn is interpreting the markdown in the summary tag properly.

anselrognlie (commented):
It looks like the two images used in Debugging Tests with Pytest under Running Tests In The Debugger should also get updated (they're still class-based)

anselrognlie (changes requested):
There are still a few things that would be nice to update. We should get the branch into a PR (I have comments related to what should be changed here in this review). And I think there are images in Debugging Tests with Pytest that should get updated.

anselrognlie (approved):
Can't put a comment since it wasn't a changed line, but in debugging-with-vs-code.md:80, can we change the line numbers to 14 and 16 to match the image?

Otherwise looks good, so signing off!

PR #128, opened 2021-05-07
typo

tgoslee:


anselrognlie (approved):
I think the it is optional here, but if it sounds better to you, then I don't see a problem with this. The space fixes are nice too.

PR #129, opened 2021-05-14
updates playlists to a playlist per page

anselrognlie:
Other learn lessons only have the videos (sometimes multiple,
if they were split into chunks) pertaining to that one lesson in the playlist
at the top of the lesson. Intro to Git had the same single playlist with all
of the videos in that single playlist on every lesson.

This change creates a separate playlist per lesson (even though there's
only one video in each lesson) and places a distinct playlist on each
lesson page.


tildeee (approved):
Thank you for making the separate playlists!!

PR #130, opened 2021-08-17
Jl/sets

jbieniosek:
This is a first draft & it's missing the vocab section.  Posting for feedback!


new_lessons/sets.md
```diff
@@ -0,0 +1,337 @@
...
+
+s = "she sells sea shells by the sea shore"
+result = []
+
+for char in s:  # loop over each character in the string
+    if char not in result: # if the character is not in the result set:
+        result.append(char) # add the character to the result set
+
+for char in result: #print out the results
+    print(char)
```
tildeee, line 287: The problem statement above doesn't mention a need for printing... and to be honest, I'm not sure if these examples need an emphasis on printing, especially via a loop? Printing a list or a set in Python usually displays in a reasonable way, IMO.

Could we work the printing out, or could we make it more meaningful?

new_lessons/sets.md
```diff
@@ -0,0 +1,337 @@
...
+
+```python
+
+s = "she sells sea shells by the sea shore"
+result_set = set(s)
+for char in result_set:
+    print(char)
+
+```
+
```
tildeee, line 319:
```suggestion

## Check for Understanding

```

tildeee: I know there's likely to be a problem set attached to sets, but I like problem sets to be... problem sets, and questions at the end of lessons in "Check for Understanding" to be on the very fundamental layer. I'd personally recommend adding MC questions to reinforce:

- How to make a set in Python
- The uniqueness of elements in a set
- What intersection does
- What union does

new_lessons/sets.md
```diff
@@ -0,0 +1,337 @@
...
+<!-- Question Takeaway -->
+<!-- prettier-ignore-start -->
+### !challenge
+* type: paragraph
+* id: 9Nqz2c
+* title: Variables Are References
```
tildeee, line 325: Need to update the ID and the title here

tildeee (commented):
In addition to these comments, the folder `new_lessons` is a lil weird. Could we pick a different topic name, and stick with kebab casing?

new_lessons/sets.md
```diff
@@ -0,0 +1,337 @@
...
+
+1. Create a new set:
+```python
+
+new_set = set() # To create an empty set, use the keyword set()
+set_with_contents = {1, 2, 3, 'a', 'b'} # To create a set with contents, use {}
+# like dictionaries.  Notice the difference from dictionaries, each element is
+# on it's own, there are no key-value pairs.
+
+```
```
tildeee, line 37: Overall throughout this lesson, there's inconsistency between if code snippets have empty lines on top and bottom, only on bottom, only on top, or none.

Could we conform to one?

The rest of the curriculum does not have empty lines on the top or bottom.

new_lessons/sets.md
```diff
@@ -0,0 +1,337 @@
...
+1. Add an element to a set:
+```python
+
+set_a = {1, 2, 3}
+set_a.add(4) # set_a now equals {1, 2, 3, 4}
+
+```
+1.  Remove an element from a set:
+* There are two ways to remove an element, remove() which raises an error if the element is not present, and discard() which does not.
```
tildeee, line 46: Why is this formatting like this?

new_lessons/sets.md
```diff
@@ -0,0 +1,337 @@
...
+set_a = set(list_a) # converts the list into a set!
+set_b = set(list_a) 
+
+result_set = set_a & set b
+
+return list(result_set)
+
+```
+
+### Iterating Through Sets
```
tildeee, line 273: Why is this section named "Iterating Through Sets"? I don't see iteration here. Also, I'm not exactly sure what I'm supposed to get out of this section, tbh... I feel like this is more of an example of converting a list to a set.

new_lessons/sets.md
```diff
@@ -0,0 +1,337 @@
...
+
+Let's say we have a string of letters and we want to find all of the unique characters in the string.  Here's a brute force solution that we could use:
+
+```python
+
+s = "she sells sea shells by the sea shore"
```
tildeee, line 279: Since there's a lot of "s" words (set... string... she sells sea shells...) could we find a different variable name besides `s`?

new_lessons/sets.md
```diff
@@ -0,0 +1,337 @@
...
+
+| Vocab | Definition | Synonyms | How to Use in a Sentence |
+| ----------------- | ----------------------------------------------------------------------------------------------------- | -------- | --------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Set |  |  |
+| Union |  |  |
+| Intersection |  |  |
+| Difference |  |  |
+| Disjoint |  |  |
+
+## Set Basics
```
tildeee, line 17: As mentioned in another comment, if there's another word for "Basics," such as "Fundamentals," or only "Sets," I think that'd be great.

tildeee (commented):
I re-read this lesson through Learn preview and added a few more comments that stood out to me.

PR #132, opened 2021-08-19
Re-org first pass

jbieniosek:
(cherry picked from commit 9fdce921f0d400c57fc7023fff9c26c4d51d5394)
Signed-off-by: Jasmine Lopez <jasmine@adadevelopersacademy.org>

Re-org pass 2


config.yaml
```diff
@@ -437,7 +453,7 @@ Standards:
       -
         Type: Lesson
         UID: ecf610c94fd56dbcc8dcfa0f774342ec
-        Path: /intro-to-object-oriented-programming/problem-set-intro-to-oop.md
+        Path: /intro-to-object-oriented-programming/problem-set-constructors.md
```
tildeee, line 456: Should this be replacing the Intro to OOP Problem Set? The description implies that there are two problem sets.

tildeee (commented):
In general, I love it and I agree with it!! But the "Representing Data" topic is now pretty big. In general, I'd feel uncomfortable asking students to absorb all of it in as part of the same topic. Could we move "Variable Scope" and "Variables Are References" into a split topic?

config.yaml
jbieniosek: Sorry, That was a mistake, there should be two PS for now (not sure if that's the end goal or if the intro to oop problem set should have a constructors section, something to discuss) based on the task airtable.  I will submit a fix for that!

jbieniosek:
I agree, it's a big topic!  Would combining Sets & Tuples into one shorter lesson and having a sets and tuples section in the problem set rather than having their own problem sets have the same shortening effect?

tildeee:
Hmmm. I like the size of your Set lesson as a single lesson. Also, I personally think it's nicer organization/findability if Sets and Tuples are separate lessons, even if they may be small/regardless of size...

So I actually think it's better off with Sets and Tuples being separate as-is! If you can reassure me that it'll be okay if students need to absorb all of that new information in one day, then I'll feel better 🥺  lololol

jbieniosek:
I don't have a strong argument for having student go over all of this in the same day.  There were only 3 topics on the Week 4, so adding another topic would fit right in and everything could shift down.  I like the idea of spending a whole day on variables under the hood, but I think if it gets broken into it's own topic it would benefit from a PS to go along with it.

tildeee (approved):
If we split the Variables into a new topic:

I'm definitely happy to pitch in and make a PS for the new topic if you'd enjoy making a placeholder for me!

If the idea of making a new PS at this moment feels too rushed, let's go with your original plan, and I'm grateful we got to explore this! I trust your decision on it.

PR #137, opened 2021-08-27
Aa/super

audreyandoy:
- On Wednesday: Ansel, Jasmine, and I met and revisited intro to oop and determined that a super() rewrite needs occur more than a constructors lesson. 
- Here is a start on super, however, I wonder if we should combine inheritance and super lesson together as a pattern we encourage for students is to always use the super constructor for child classes.


anselrognlie (commented):
some rewording in the inheritance lesson to help clarify what we're looking forward to. I have an additional PR against this branch for the super portion.

PR #139, opened 2021-08-28
modifications to de-emphasize multiple inheritance

anselrognlie:
I'd _really_ like us to stay away from multiple-inheritance discussions. Even with Python's official support for it (and mixins) it's still non-trivial to get right, and not something a junior dev needs to be thinking about.

I reworked the sample code to show reasons to use super in constructors as well as regular instance methods. I added some CFU questions that reinforce the ctor usage, and discourages using super when it's not needed.

I think this lesson provides enough of a justification for using super (especially if we avoid introducing some of the alternatives), so hopefully this is helpful!

### DIFF
--- a/intro-to-oop-relationships/super.md
+++ b/intro-to-oop-relationships/super.md
@@ -1,50 +1,20 @@
 # super()
 
 ## Learning Goals
+
 - Use super() in inheritance
 
-## Inheritance and constructors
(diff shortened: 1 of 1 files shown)


jbieniosek (approved):

audreyandoy (approved):
LGTM 👍

PR #145, opened 2021-09-22
Change 4.67 to 4

beccaelenzil:


intro-to-object-oriented-programming/defining-classes.md
```diff
@@ -454,11 +454,11 @@ from main import *
 
 class TestDriverClass(unittest.TestCase):
   def test_driver_has_attributes(self):
-    speed_racer = Driver("DR0004", 35, [], 4.67)
```
anselrognlie, line 457: What's the reason for this? A float should (in principal) work fine. Though this looks like another test where `assertEqual` should be being used throughout rather than `assertIs`. We really care about value equality here (`assertEqual`) rather than object identity (`assertIs`).

anselrognlie (commented):
What's the motivation for this change? It doesn't seem like we should need to do this (though we might want to change the asserts being checked).

PR #163, opened 2022-02-28
Update challenge 2 in testing objects

beccaelenzil:
Cherise noticed that the answers for challenge two were incorrect/unclear. I put in her corrections.


cheriseatada (approved):
Looks good

PR #167, opened 2022-03-10
Add space to print stmt

CheezItMan:
We could also use regex in the answer.


anselrognlie (approved):
Saw the comment about this in Learn. Looks good!

beccaelenzil (approved):

PR #173, opened 2022-03-17
For instructors: example edge cases for Blackjack TDD Exercise

goeunpark:
## Context
Some students finished the Blackjack TDD exercise much earlier than time allotted and instructors didn't have a go-to backup task for them. 

## Offer
Changes here are only for the instructor-facing TDD Exercise description page. These are additional edge tests we can offer students. Goeun's lukewarm take: Scoring two aces in blackjack is a challenging enough problem that it shouldn't be included in the main exercise but for folks who feel comfortable with loops and conditionals, it can keep them occupied for another 30min - hour.


anselrognlie (approved):
LGTM. Always nice to have extra material on hand if needed.

beccaelenzil:
Here's a suggestion for adding these edge cases to the test file: https://github.com/AdaGold/tdd-exercise/pull/6

PR #175, opened 2022-04-01
fix typo, Bear -> Mammal

beccaelenzil:
To be merged after Unit 1 completes and classes are switched to C17 branches


beccaelenzil:
This should not be merged in. I accidentally removed the third correct solution, so we should restore that third, correct solution.

PR #176, opened 2022-04-05
Fix small typo in last challenge of problemset

kelsey-steven-ada:
`Bedroom` is styled like a class name in the challenge description, but the class in the code is named `Room`. Update description to match the code.

### DIFF
--- a/composition/problem-set-composition.md
+++ b/composition/problem-set-composition.md
@@ -99,7 +99,7 @@ print(today.display_temperature())
 * title: Composition
 
 ##### !question
-Composition can be initiated by a single line of code. In this code snippet, which line is responsible for starting the relationship between `Bedroom` and `Area`?
+Composition can be initiated by a single line of code. In this code snippet, which line is responsible for starting the relationship between `Room` and `Area`?
 
 ```python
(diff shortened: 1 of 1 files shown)


composition/problem-set-composition.md
```diff
@@ -99,7 +99,7 @@ print(today.display_temperature())
 * title: Composition
 
 ##### !question
-Composition can be initiated by a single line of code. In this code snippet, which line is responsible for starting the relationship between `Bedroom` and `Area`?
+Composition can be initiated by a single line of code. In this code snippet, which line is responsible for starting the relationship between `Room` and `Area`?
```
anselrognlie, line 102: _Should_ this be `Room`? Or should it be `bedroom`? Definitely not `Bedroom`, though... 😂

kelsey-steven-ada: I suppose the answer is the same either way 😆 . 

@beccaelenzil Would you be the right person to ask for the original intent?

beccaelenzil: I don't know what the original intent was, but `Room` and `Area` sounds like a good fix to me!

PR #184, opened 2022-07-22
add unit 1 sequence instructor doc

beccaelenzil:


fundamentals/sequence.instructor.md
```diff
@@ -0,0 +1,39 @@
+# Instructor: Unit 1 Topic Sequence
+
+* Unit 1: Fundamentals
```
kelsey-steven-ada, line 3: For my knowledge, is it placed in Fundamentals since that's always the first tile so it's at the "start" of the Unit 1 content? It feels a little weird to me for this doc to live inside of any given lesson, but I can't think of other suggestions if we want it to be in Learn.

kelsey-steven-ada (approved):

fundamentals/sequence.instructor.md
beccaelenzil: That's why I placed it there. We can't make a topic that just has Instructor docs (which would be my preference). Do you think it would make more sense to have it in the About Ada Topic in the About Ada Unit?

kelsey-steven-ada: I think in the About Ada Topic makes a lot of sense if we plan to have sequence docs for each unit there =]

PR #191, opened 2022-09-08
Higher Order Functions Topic

beccaelenzil:
- The problem set is relatively short. There are a few check for understanding questions in the lessons.
    - Question: do you think we need more questions?
    - suggestions are welcome for problem set questions. Or maybe you want to write one or two :) 
- I proofread each lesson and confirmed in a python script that all the code samples work (there were many errors the first time through :)
   - I anticipate a bit more proofreading will still be required.
   - code samples python scripts do not need to be reviewed. These are not published. They are used to confirm the code samples work and are consistently formatted. Feel free to use them to test code samples.
- Feedback focused on clarity of explanations is much appreciated (and anything else)
    - suggestions on rewording is most welcome
Can be previewed in Core Sandbox or using Learn preview: https://learn-2.galvanize.com/cohorts/3200/blocks/1032/content_files/higher-order-functions/min-max-sorted.md


anselrognlie (changes requested):
I went through a pretty holistic editing pass on this, so it didn't fit well in individual comments. Github just isn't great for collaboratively editing plain text.

Instead, I created a follow on [PR](https://github.com/Ada-Developers-Academy/core-unit-1/pull/192) against this PR branch (#192) with all my feedback applied.

If you like all my suggestions, it's easy to pull it all in by merging. Unfortunately, if you want to bring in the changes piecemeal, it requires copying sections across manually. But I still think this is the most convenient way to deliver my feedback.

Here's an overview of the changes I made:

- General rewording, with an emphasis on flow, parallel construction of clauses, and expansion of explanations
- Wrapped the three sample functions in min/max/sorted in summary views to encourage thinking through an implementation
- Added missing spaces in commented code samples to keep typical alignments
- Introduced the vocab names parameter as a synonym for keyword argument
- Inserted some zero-width spaces to allow long code in tables to wrap more naturally
- Renamed example `sum` function to `add` to avoid redefining the built-in function, but still keep the name short for the explanation table
- Adjusted table column widths for readability
- Added reference to PEP 3102 for students to read about functions that require a parameter to be a keyword argument (like the key in the functions we're looking at)
- Expanded details about why we can reassign functions to variables and use them
- Expanded description of lambda syntax
- Expanded description of when to use lambdas
- Fixed explanation for first lambda challenge (was copy/pasted second explanation)
- Got a little wishy-washy around the use of callable in the second lambda challenge explanation, since we haven't really introduced that concept
- Fixed some mismatched explanation in the generalization of get_first_movie to min_function_custom
- Renamed get_value_for_key parameter to get_value_from_item to better reflect the parameters those functions take
- Added callout addressing possible concerns that this refactor is overkill for just changing which key we are using
- Formatting updates for many of the code blocks in writing higher order functions
- Changed final problem set challenge (sorting) to not have the special `None` behavior for an empty list. Calling a sort method should always return a list, and changing the return type for an empty list doesn't sit well with me.

higher-order-functions/min-max-sorted.md
```diff
@@ -0,0 +1,320 @@
+# Passing Functions as Arguments: Min, Max, and Sort
+
+## Learning Goals
+
+* Demonstrate how to pass a function as an argument
+* Describe how to use the key parameter in the `sorted` function to sort a list of dictionaries by a specific key
+* Describe how to use the key parameter in the `min` function to find the minimum in a list of dictionaries by a specific key
```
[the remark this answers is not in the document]
kelsey-steven-ada: I had a similar thought. I was wondering if it would it be worth showing an example with lists (like sorting by title length) to explicitly show data structures other than dictionaries can provide custom comparison keys.

higher-order-functions/writing-higher-order-functions.md
```diff
@@ -0,0 +1,171 @@
...
+    for movie in movies:
+        if get_title(movie) < min_title:
+            min_movie = movie
+            min_title = get_title(movie)
+    return min_movie
+```
+
+Now, let's take it one step further to generalize our function to find the minimum movie by any key (i.e. `"title"`, `"rating"`, or `"release_year"`). 
+
+We will add a second parameter, a function that gets the value for a specified key: `get_value_for_key`. We will generalize our variable names changing `movies` to `collection` and `movie` to `item.` Each line from `get_first_movie` that has the expression `get_title(movie)` will be replaced with `get_value_for_key(item)`.
```
kelsey-steven-ada, line 106: Though it's not, for me, at first it felt like noticeably more changes on this step vs the previous. It might be more digestible as a slightly re-ordered list, something like:
```suggestion
We will: 
1. generalize our variable names by changing `movies` to `collection` and `movie` to `item.`
2. add a second parameter, a function that gets the value for a specified key: `get_value_for_key`.
3.  replace all uses of the expression `get_title(movie)` with a call to our new parameter `get_value_for_key(item)`.
```

higher-order-functions/writing-higher-order-functions.md
```diff
@@ -0,0 +1,171 @@
...
+    min_movie = movies[0]
+    min_title = get_title(movies[0])
+    for movie in movies:
+        if get_title(movie) < min_title:
+            min_movie = movie
+            min_title = get_title(movie)
+    return min_movie
+```
+
+Now, let's take it one step further to generalize our function to find the minimum movie by any key (i.e. `"title"`, `"rating"`, or `"release_year"`).
```
kelsey-steven-ada, line 104: We talk about finding the minimum of a movie based on attributes on this line, but the the description of the step removes all movie-related naming. I'm wondering if it's worth splitting out the renaming as a final step - something like "We now have a flexible custom min function for movies, but this would work for any list of dictionaries. Let's rename our variables for clarity!"

higher-order-functions/writing-higher-order-functions.md
```diff
@@ -0,0 +1,171 @@
...
+```py
+first_movie_by_title = min_function_custom(
+    movies, get_value_for_key=lambda movie: movie["title"])
+lowest_rated_movie = min_function_custom(
+    movies, get_value_for_key=lambda movie: movie["rating"])
+earliest_movie = min_function_custom(
+    movies, get_value_for_key=lambda movie: movie["release_year"])
+```
+
+We can also pass the second parameter as a positional argument rather than a keyword argument:
```
kelsey-steven-ada, line 147: I might have missed it, but I don't recall seeing an example of moving the order of arguments with keyword arguments here or in the previous lesson that introduced them. I think it'd be nice to have a callout around "keyword arguments can be passed in a different order from how they're defined, but that cannot be done with positional"

higher-order-functions/min-max-sorted.md
```diff
@@ -0,0 +1,342 @@
...
+The following table describes valid and invalid syntax for calling `get_the_sum`:
+
+| Function Call | Valid Syntax (yes/no) | Explanation |
+|--|--|--|
+|`sum(1, 2)` | yes | This is an example of calling a function with two positional arguments. |
+|`sum(a=1, b=2)` | yes | This is an example of calling a function with two keyword arguments. |
+|`sum(1, b=2)` | yes | This is an example of calling a function with a mix of positional and keyword arguments. |
+|`sum(a=1, 2)` | no | This is invalid. Positional arguments cannot follow keyword arguments. <br> `SyntaxError: positional argument follows keyword argument`|
+
+## Higher Order Functions
```
kelsey-steven-ada, line 267: It looks like the topic is named "Higher Order Functions", but we don't define that term until this section. I'm wondering if it could be helpful to add a definitions table to surface the terms "Higher Order Functions", "keyword argument", and "positional argument".

higher-order-functions/min-max-sorted.md
```diff
@@ -0,0 +1,342 @@
...
+
+A higher order function is a function that takes another function as an argument.
+
+### !end-callout
+
+A function is an object in memory that is a set of instructions. A function name is a variable that points to this set of instructions. 
+
+When we use the `()` syntax, we are calling the address of that variable. When we leave off the `()`, we are looking up the address that the variable is pointing to.
+
+Just like variables that point to other objects (strings, integers, lists, etc.), we can assign a function to another variable:
```
kelsey-steven-ada, line 283: I don't think it's _necessary_ for understanding, but we could add a little clarity about where `get_title` is coming from in the example below 

```suggestion
Just like variables that point to other objects (strings, integers, lists, etc.), we can assign a function to another variable. Assuming `get_title` is a function as defined in the example above, we could write:
```

higher-order-functions/min-max-sorted.md
```diff
@@ -0,0 +1,342 @@
...
+
+# movie_titles = [
+#   '2 JavaScript 2 React', 
+#   'Javascript 3: VS Code Lint', 
+#   'Recursion',
+#   'Zero Dark Python'
+# ]
+```
+
+Alteratively, we can use the function `sorted` to return the list sorted alphabetically:
```
kelsey-steven-ada, line 45: Since we're explicit about the `sort` method sorting a list in-place, we may want to be explicit that `sorted` returns a new array:

```suggestion
Alternatively, we can use the function `sorted` to return a copy of the list sorted alphabetically:
```

higher-order-functions/min-max-sorted.md
```diff
@@ -0,0 +1,342 @@
...
+
+* Demonstrate how to pass a function as an argument
+* Describe how to use the key parameter in the `sorted` function to sort a list of dictionaries by a specific key
+* Describe how to use the key parameter in the `min` function to find the minimum in a list of dictionaries by a specific key
+
+## Introduction
+
+Imagine we have a list of movie titles we'd like to sort alphabetically. Python provides a couple convenient functions for this task. The [`sort`](https://docs.python.org/3/howto/sorting.html#sorting-basics) method can be called on a list and sorts the list in place. The `sorted` function takes a list as a parameter and returns the sorted list.  
+
+But what if we have a list of dictionaries representing movies that we want to sort by title, or perhaps by rating. Or perhaps we want to find the most recent movie or oldest movie in a list.
```
kelsey-steven-ada, line 13: The "or perhaps" repeated close together feels a little weird, I'd consider replacing one of them with "maybe" or "possibly".

higher-order-functions/min-max-sorted.md
```diff
@@ -0,0 +1,342 @@
...
+
+## Learning Goals
+
+* Demonstrate how to pass a function as an argument
+* Describe how to use the key parameter in the `sorted` function to sort a list of dictionaries by a specific key
+* Describe how to use the key parameter in the `min` function to find the minimum in a list of dictionaries by a specific key
+
+## Introduction
+
+Imagine we have a list of movie titles we'd like to sort alphabetically. Python provides a couple convenient functions for this task. The [`sort`](https://docs.python.org/3/howto/sorting.html#sorting-basics) method can be called on a list and sorts the list in place. The `sorted` function takes a list as a parameter and returns the sorted list.
```
kelsey-steven-ada, line 11: "The `sorted` function takes a list as a parameter and returns the sorted list."
- It might be worth being specific that `sorted` accepts any iterable data type, since that is significantly different from `sort` as a list instance method.

kelsey-steven-ada (commented):
This lesson is going to be so helpful! Most of my feedback is small suggestions or questions for clarity; I tried to make sure I didn't duplicate feedback from Ansel. ^_^

PR #192, opened 2022-09-11
Feedback for PR #191 (higher order functions)

anselrognlie:
- General rewording, with an emphasis on flow, parallel construction of clauses, and expansion of explanations
- Wrapped the three sample functions in min/max/sorted in summary views to encourage thinking through an implementation
- Added missing spaces in commented code samples to keep typical alignments
- Introduced the vocab names parameter as a synonym for keyword argument
- Inserted some zero-width spaces to allow long code in tables to wrap more naturally
- Renamed example `sum` function to `add` to avoid redefining the built-in function, but still keep the name short for the explanation table
- Adjusted table column widths for readability
- Added reference to PEP 3102 for students to read about functions that require a parameter to be a keyword argument (like the key in the functions we're looking at)
- Expanded details about why we can reassign functions to variables and use them
- Expanded description of lambda syntax
- Expanded description of when to use lambdas
- Fixed explanation for first lambda challenge (was copy/pasted second explanation)
- Got a little wishy-washy around the use of callable in the second lambda challenge explanation, since we haven't really introduced that concept
- Fixed some mismatched explanation in the generalization of get_first_movie to min_function_custom
- Renamed get_value_for_key parameter to get_value_from_item to better reflect the parameters those functions take
- Added callout addressing possible concerns that this refactor is overkill for just changing which key we are using
- Formatting updates for many of the code blocks in writing higher order functions
- Changed final problem set challenge (sorting) to not have the special `None` behavior for an empty list. Calling a sort method should always return a list, and changing the return type for an empty list doesn't sit well with me.


beccaelenzil:
Thank you for this thorough review. This all look good to me. I will wait for @kelsey-steven-ada's review before merging in.

PR #193, opened 2022-09-12
add a sort list of strings by length example

beccaelenzil:


anselrognlie (approved):
Nice! Thanks for providing a non-key-based example!

kelsey-steven-ada (approved):
Thanks for adding this!

PR #196, opened 2023-02-09
Adding extra context to problem 14, to make it less of a trick question

chimerror:
Fixing [this Asana Issue](https://app.asana.com/0/1201700438643002/1202980923319137/f).

Problem 14 in the Intro to Tests problem set is a bit of a trick question as it stands since you do not need to do a full implementation of determining if the input is odd, simply enough of an implementation for the test to pass.

I added a bit more of a hint to the question to write a _minimal_ implementation, as well as explaining why the answer is so minimal.


nancy-harris (approved):
LGTM

PR #198, opened 2023-02-09
Replace problems in iterating over data problem set.

alope107:
- Deleted two problems that asked students to iterate over a dictionary to find the value associated with a key.

- Added a new problem that requires dictionary iteration but does not promote poor coding practices.


anselrognlie (commented):
LGTM! 👍 

I like this MUCH better.

anselrognlie (approved):
Whoops. Meant to approve.

PR #203, opened 2023-03-10
Add callout about ordered dicts

kyra-at-ada:
**Summary of Changes**
Add callout below table on dictionaries to highlight that dictionaries are maintained in insertion order, but folks should still generally treat them as unordered.

### DIFF
--- a/representing-data/representing-data-in-lists-and-dictionaries.md
+++ b/representing-data/representing-data-in-lists-and-dictionaries.md
@@ -123,6 +123,8 @@ Dictionaries are great at...
 ## *Dictionaries & Ordering Data
 In recent versions of Python, key-value pairs in dictionaries are maintained in insertion order meaning that key-value pairs are organized in the order that they were added to the dictionary. Ordered dictionaries were first introduced as an implementation detail in version [Python 3.6](https://docs.python.org/3/whatsnew/3.6.html#new-dict-implementation) and became an official language feature beginning with [version 3.7](https://docs.python.org/3.7/library/stdtypes.html#mapping-types-dict).
 
+<br>
+
 Despite these changes, we recommend still treating dictionaries as unordered. Because dictionaries are not ordered alphabetically or numerically, treating dictionaries as ordered may not be useful or reliable unless we know how items were added to the dictionary. Additionally, similar data types in other programming languages are often unordered. 
 
 ### !end-callout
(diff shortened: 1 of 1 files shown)


representing-data/representing-data-in-lists-and-dictionaries.md
```diff
@@ -114,9 +114,19 @@ Dictionaries are great at...
...
 | Accessing values by using keys            | All values can be accessed by a key; this might have more meaning than accessing values by index, especially when keys are named intentionally                                                      | We can access values by using square brackets, ```dict_name["key_name"]```    |
 
+<!-- available callout types: info, success, warning, danger, secondary, star  -->
+### !callout-star
+
+## *Dictionaries & Ordering Data
+In recent versions of Python, key-value pairs in dictionaries are maintained in insertion order meaning that key-value pairs are organized in the order that they were added to the dictionary. Ordered dictionaries were first introduced as an implementation detail in version [Python 3.6](https://docs.python.org/3/whatsnew/3.6.html#new-dict-implementation) and became an official language feature beginning with [version 3.7](https://docs.python.org/3.7/library/stdtypes.html#mapping-types-dict).
+
```
anselrognlie, line 125: Include a break so learn doesn't squash the paragraphs
```suggestion

<br>

```

representing-data/representing-data-in-lists-and-dictionaries.md
```diff
@@ -114,9 +114,19 @@ Dictionaries are great at...
...
+| Data is unordered*                         | Dictionaries may sometimes appear to have an order, but there is no consistent ordering of key-value pairs. This is helpful to express that there isn't necessarily priority or importance to order | ``` {"key_123": 10,    "key_200": 0.34,   "key_2": 1}``` Notice how despite there being a numbering convention in these key-value pairs that they are not in a specific order.  |
 | Accessing values by using keys            | All values can be accessed by a key; this might have more meaning than accessing values by index, especially when keys are named intentionally                                                      | We can access values by using square brackets, ```dict_name["key_name"]```    |
 
+<!-- available callout types: info, success, warning, danger, secondary, star  -->
+### !callout-star
```
anselrognlie, line 121: I didn't know about this callout style!

Since it ends up adding a star icon, could we get by without the asterisk below? I can go either way.

char-adadev (approved):
Looks good to go, thank you for adding this callout!

anselrognlie (approved):
Looks good. The text is really clear. Just some minor formatting feedback.

representing-data/representing-data-in-lists-and-dictionaries.md
kyra-at-ada: That was my original plan but it's a different kind of a star. It's a 5 pointed star. I could change it back to 'info' to make it less redundant

anselrognlie: LOL. I tried sticking the same star on the initial text and unfortunately it looks ridiculous.

As an alernative `callout-` with an "invalid" style (like none, so `callout-none`) still makes a callout, but leaves of the glyph.

PR #208, opened 2024-08-19
Remove Global Installation  References from Unit 1

mikellewade:
[Asana Ticket](https://app.asana.com/0/1201700438643002/1206923601980908/f)


intro-to-tests/fizz_buzz.instructor.md
```diff
@@ -11,8 +11,7 @@ The live code repo: [TDD Ping Pong /Fizz Buzz](https://github.com/AdaGold/tdd-pi
 The suggested way to run this live code is one instructor writes a test and then the other instructor implements code to make the test pass. The second instructor then writes the next test.
 
 A `requirements.txt` is not currently provided. Here are a few suggestions for how to run this exercise related to the `requirements.txt`.
-- We could not use venv and use the globally installed pytest for this exercise
 - We could use venv and install pytest and create a requirements.txt
 - We could create a `requirements.txt` in the repo and model creating a venv and installing the requirements
 
-In C15 and C16 we introduced `requirements.txt` and `venv` with the first project.
\ No newline at end of file
+In C15 and C16 we introduced `requirements.txt` and `venv` with the first project. Starting with C21, introduced `requirements.txt` and `venv` in precourse due to Python no longer allowing the global installation of packages.
```
kelsey-steven-ada, line 17: I'm wondering if this historical context is useful in the resources for the activity? If others are okay with it, I'm good removing this chunk-we have the git commit history if we need to look at when we made changes to the lesson.

```suggestion
```

intro-to-tests/intro-to-pytest.md
```diff
@@ -21,10 +22,30 @@ When we understand that tests can verify code correctness, let's learn one way t
 
 [`pytest`](https://docs.pytest.org/en/stable/) is one of many testing tools for the Python language.
 
-We can install (and upgrade) the `pytest` package with this command:
+
+
+Before we install `pytest`, let's recall the steps we took in Precourse the commands to install Python packages. First, we need to navigate to the directory that we want to setup our virtual environment in. We can use the `pwd` command to check if we are currently inside of the desired directory, if not then we can use the `cd` command to navigate to the correct one. After we are inside the correct directory we can the following commands to create and then activate our virtual environment:
```
kelsey-steven-ada, line 27:
```suggestion
Before we install `pytest`, let's recall the steps we took in Precourse the commands to install Python packages:
1. First, we need to navigate to the directory that we want to setup our virtual environment in. 
    - We can use the `pwd` command to check if we are currently inside of the desired directory. 
    - If we're not in the right folder, then we can use the `cd` command to navigate to the correct one. 
2. After we are inside the correct directory we can use the following commands to create and then activate our virtual environment:
```

intro-to-tests/intro-to-pytest.md
```diff
@@ -21,10 +22,30 @@ When we understand that tests can verify code correctness, let's learn one way t
...
+$ source venv/bin/activate
+```
+
+Remember that the first command creates a virtual environment under the name `venv` and the second command activates our virtual environment. After running the activation command, we should now see that each line of our terminal now begins with `(venv)`. This lets us know that our virtual environment has been activated. 
+
+### !callout-info
+
+## Virtual Environments
+
+At high level, virtual environments are used to  install and manage packages separately from other projects that could be using the same packages. For instance, if project A requires an older version of `pytest` than project B, using a virtual environment allows both projects to exist on the same machine without have to upgrade/downgrade `pytest` to meet the conflicting version requirements. We will discuss virtual environments in greater detail later in `Managing Packages` and when discussing the `Adagrams` project.
```
kelsey-steven-ada, line 40:
```suggestion
At high level, virtual environments are used to install and manage packages separately from other projects that could be using the same packages. For instance, if project A requires an older version of `pytest` than project B, using a virtual environment allows both projects to exist on the same machine without needing to upgrade/downgrade `pytest` to meet the conflicting version requirements. We will discuss virtual environments in greater detail later in `Managing Packages` and when discussing the `Adagrams` project.
```

intro-to-tests/intro-to-pytest.md
```diff
@@ -60,11 +81,12 @@ Take care to notice the following:
    - This represents a test file that is responsible for testing `projectname/somefile.py`
 1. There are some other files: `__init__.py` and `context.py`
```
kelsey-steven-ada, line 82: GitHub won't let me select the first bullet on line 55 to suggest a change, but we should update it to acknowledge there's also a `venv` folder as well.

```suggestion
1. There are some other files such as `__init__.py`
```

intro-to-tests/intro-to-pytest.md
```diff
@@ -21,10 +22,30 @@ When we understand that tests can verify code correctness, let's learn one way t
...
+
+
+Now let's use the command below to install `pytest`:
 
 ```bash
-pip3 install -U pytest
+(venv) pip3 install -U pytest
 ```
 
 ### !callout-danger
```
kelsey-steven-ada, line 51: To me, this doesn't feel like it fits a `callout-danger` since it's reminding folks about something they've already seen. Would you be able to swap this over to an info callout while in the file?
```suggestion
### !callout-info
```

intro-to-tests/fizz_buzz.instructor.md
mikellewade: Yeah, I'm fine with that. Plus, we have it in Asana as well in the curriculum tasks. I can remove it.

intro-to-tests/intro-to-pytest.md
mikellewade: Sorry, could you explain what you mean by "...while in the file"? I want to make sure I'm not missing anything.

kelsey-steven-ada: All good! I was checking if you were comfortable to make this update while you're already making changes to this lesson, even though this particular change is outside the scope of the global installation references work you were doing.

mikellewade: Oh! Yes, that's fine by me!

intro-to-tests/intro-to-pytest.md
```diff
@@ -75,15 +75,15 @@ When we use `pytest`, we will be creating more files. In order to anticipate mor
...
-1. In our project directory, there are 0 files, and 2 folders: `projectname` and `tests`.
+1. In our project directory, there are 0 files, and 3 folders: `projectname`, `tests`, and `venv`.
    - We will replace the folder name `projectname` with our project's name
-1. In our `projectname` folder, there is a file named `somefile.py`
+2. In our `projectname` folder, there is a file named `somefile.py`
    - This represents any Python files that include the functions we're writing
-1. In our `tests` folder, there is a file named `test_somefile.py`
+3. In our `tests` folder, there is a file named `test_somefile.py`
    - This represents a test file that is responsible for testing `projectname/somefile.py`
 1. There are some other files such as `__init__.py`
```
kelsey-steven-ada, line 84: Since the others bullets moved to sequential numbers, I think we need to update the number here as well for the order to flow right.
```suggestion
4. There are some other files such as `__init__.py`
```

kelsey-steven-ada (commented):
Looking good! Only thing I see is one bullet that didn't move to the sequential numbering.

intro-to-tests/intro-to-pytest.md
mikellewade: Oops! Let me fix that!

kelsey-steven-ada (approved):

PR #214, opened 2025-01-30
Add hints and explanations to questions 19-22

yangashley:
[Explanations for Problem Set: Fundamentals Vocabulary](https://app.asana.com/0/1205655776549324/1204113279723331/f)

### DIFF
--- a/fundamentals/problem-set-fundamentals-vocabulary.md
+++ b/fundamentals/problem-set-fundamentals-vocabulary.md
@@ -559,7 +559,7 @@ A literal refers to the value itself without referencing a variable.
 ##### !explanation
 
 - The first two options are dictionary literals because they are data structures with key-value pairs that are surrounded by curly braces. They are not referenced by any variables. 
-- The third option is not valid Python syntax.
+- The third option does not contain curly braces and is therefore not valid Python syntax.
 
 ##### !end-explanation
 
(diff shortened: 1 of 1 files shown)


fundamentals/problem-set-fundamentals-vocabulary.md
```diff
@@ -617,6 +639,13 @@ Which of the following are literals?
 * `[]`
 
 ##### !end-answer
+##### !explanation
```
yangashley, line 642: I didn't add a hint here because this question is the same as the question above but with different options. I can add the same hint as what is shown above on line 599 if it feels like this question should also have a hint.

fundamentals/problem-set-fundamentals-vocabulary.md
```diff
@@ -549,6 +549,20 @@ Which of the following options is a dictionary literal?
...
+
+Dictionaries in Python are data structure that allow us to store key-value pairs. One of the ways to create a dictionary is to place key-value pairs within curly braces `{}`.
+
+A literal refers to the value itself without referencing a variable. 
+
+##### !end-hint
+##### !explanation
+
+- The first two options are dictionary literals because they are data structures with key-value pairs that are surrounded by curly braces. They are not referenced by any variables. 
+- The third option is not valid Python syntax.
```
apradoada, line 562:
```suggestion
- The third option does not contain curly braces and is therefore not valid Python syntax.
```
Maybe just a small change here to indicate why it's not valid Python syntax?

fundamentals/problem-set-fundamentals-vocabulary.md
apradoada: I think the hint from above is enough to suffice!

fundamentals/problem-set-fundamentals-vocabulary.md
```diff
@@ -617,6 +639,13 @@ Which of the following are literals?
 * `[]`
 
 ##### !end-answer
+##### !explanation
+
+- `87` is a integer literal, `"87"` is a string literal, `True` is a Boolean literal, and `[]` is a empty list literal.
```
apradoada, line 644:
```suggestion
- `87` is an integer literal, `"87"` is a string literal, `True` is a Boolean literal, and `[]` is an empty list literal.
```

fundamentals/problem-set-fundamentals-vocabulary.md
```diff
@@ -580,9 +594,17 @@ Which of the following are literals?
...
 ##### !end-answer
-##### !explanation
+##### !hint
 
+A literal value is a value itself. The value is not referenced by a variable. For example, `3` is a literal integer but `a_number = 3` is not a literal integer.
 
+##### !end-hint
+##### !explanation
+
+- `"Welcome! Make yourselves comfortable!"` is a string literal, `["magenta", "purple", "blue"]` is a list literal, and `None` is a special literal value that represents the absence of a value. These three values are actual values that we might directly write into code.
```
apradoada, line 604:
```suggestion
- `"Welcome! Make yourselves comfortable!"` is a string literal, `["magenta", "purple", "blue"]` is a list literal, and `None` is a special literal that represents the absence of a value. These three values are actual values that we might directly write into code. 
```
Just a small suggestion to avoid repetition.

apradoada (commented):
I think the formatting looks good. I like having the explanations as bullet points! I think it works well and is easy to read. I just made a few small suggestions in terms of content!

apradoada (approved):

